PR #10, opened 2022-04-29
Widget UI improvements

operramon:


libs/search-widget/src/components/modal/Modal.svelte
```diff
@@ -100,11 +100,25 @@
...
 
+  .modal.transparent .modal-content {
+    overflow: auto;
+    padding-right: 16px;
+  }
+  .modal.transparent .modal-content::-webkit-scrollbar {
+    width: 6px;
+  }
+  .modal.transparent .modal-content::-webkit-scrollbar-thumb {
+    background-color: #000;
```
ebrehault, line 116: use a `var` (define new ones in `wdiget.svelte` if this color is not existing yet)

libs/search-widget/src/components/modal/Modal.svelte
```diff
@@ -100,11 +100,25 @@
...
+  }
+  .modal.transparent .modal-content::-webkit-scrollbar-track {
+    background-color: #fff;
```
ebrehault, line 119: use a `var` (define new ones in `wdiget.svelte` if this color is not existing yet)

libs/search-widget/src/widgets/row/Row.svelte
```diff
@@ -65,19 +65,25 @@
...
   }
   .title {
+    margin-bottom: 0.4em;
     font-weight: var(--font-weight-bold);
   }
+  .byline {
+    font-size: 12px;
+    font-style: italic;
+    padding-bottom: 10px;
+    border-bottom: 1px solid #000;
```
ebrehault, line 85: use a `var` (define new ones in `wdiget.svelte` if this color is not existing yet)

libs/search-widget/src/widgets/suggestions/Suggestions.svelte
```diff
@@ -60,5 +60,6 @@
     text-align: right;
     font-size: small;
     margin: 0;
+    color: #707070;
```
ebrehault, line 63: use a `var` (define new ones in `wdiget.svelte` if this color is not existing yet)

ebrehault (approved):

libs/search-widget/src/components/modal/Modal.svelte
operramon: fixed

libs/search-widget/src/components/modal/Modal.svelte
operramon: fixed

libs/search-widget/src/widgets/row/Row.svelte
operramon: fixed

PR #24, opened 2022-05-10
Small issues

operramon:


apps/app/src/app/account/account-dbs/account-dbs.component.ts
```diff
@@ -12,43 +11,55 @@ import { TokenDialogComponent } from '../../components/token-dialog/token-dialog
...
+  constructor(
+    private stateService: StateService,
+    private nucliaDBService: NucliaDBService,
+    private dialog: MatDialog,
+    private cdr: ChangeDetectorRef
+  ) { }
 
   ngOnInit(): void {
-    this.updateKeys();
+    this.updateKeys().subscribe();
```
ebrehault, line 40: it would be safer to have a `takeUntil` here

ebrehault (approved):

apps/app/src/app/account/account-dbs/account-dbs.component.ts
operramon: Fixed

PR #41, opened 2022-05-23
desktop UI: steps, files confirmation and topbar

operramon:


apps/desktop/src/app/main-layout/topbar.component.ts
```diff
@@ -1,16 +1,46 @@
...
+    private router: Router,
+    private user: UserService,
+    private sync: SyncService,
+    private sdk: SDKService,
+  ) {}
+
+  menuOpen = false;
+
+  account = of(this.sync.getAccount()).pipe(
+    switchMap((account) => this.sdk.setCurrentAccount(account)),
```
ebrehault, line 27: Why do we need to do that here?
It is already done in the SyncService constructor.
(I mean, maybe we need to repeat the `setCurrentAccount` for some reason, but if so, that's fine, even though it is a bit weird. Maybe at some point we will need a `getCurrentAccount` in SDKService)

apps/desktop/src/app/upload/steps/steps.component.ts
```diff
@@ -0,0 +1,17 @@
+import { Component, Input } from '@angular/core';
+
+@Component({
+  selector: 'da-steps',
+  templateUrl: './steps.component.html',
+  styleUrls: ['./steps.component.scss']
```
ebrehault, line 6: OnPush

ebrehault (approved):

apps/desktop/src/app/upload/steps/steps.component.ts
operramon: fixed

apps/desktop/src/app/main-layout/topbar.component.ts
operramon: fixed

PR #61, opened 2022-06-07
display splash screen before redirecting to dashboard

ebrehault:


apps/desktop/src/app/login/login.component.scss
```diff
@@ -0,0 +1,24 @@
+@import 'variables';
+.splash {
+  position: fixed;
+  top: 0;
+  left: 0;
+  height: 100%;
+  width: 100%;
+  z-index: 99999;
```
mpellerin42, line 8: is this super high z-index really needed?

apps/desktop/src/app/login/login.component.scss
```diff
@@ -0,0 +1,24 @@
...
+}
+
+.container {
+  position: relative;
+  top: 10vh;
+  width: 600px;
+  height: 80vh;
+  margin: 0 auto;
+  padding: 62px 100px;
+  border-radius: 8px;
```
mpellerin42, line 19: Are we using rhythm in this project?

apps/desktop/src/app/login/login.component.scss
ebrehault: that's the value we use for splash screen in our different index.html (that's why there is no SCSS variable for it, but I'll create one)

libs/pastanaga/src/lib/styles/partials/_variables.scss
```diff
@@ -254,7 +259,7 @@ $font-size: (
   // ~36px - big article promo title
   xxd: var(--stf-font-size-xxd),
   // ~42px - big section primi title
-  xxxd: var(--stf-font-size-xxxd) // ~60px - main hero claim,
+  xxxd: var(--stf-font-size-xxxd) // ~60px - main hero claim,,
```
mpellerin42, line 262: not sure those extra comma are needed 😅

mpellerin42 (approved):
Looks good to me (even though you added some commas in a comments ;)

PR #64, opened 2022-06-08
Ericbrehault/sc 1472/trigger dropbox oauth when third party token

ebrehault:


apps/desktop/src/app/sync/sync.service.ts
```diff
@@ -91,55 +106,91 @@ export class SyncService {
...
+  authenticateAllPendingConnectors(): Observable<void> {
+    const sources = this._queue
+      .filter((sync) => !sync.started && !sync.completed)
+      .reduce((acc, sync) => {
+        if (!acc[sync.source]) {
+          acc[sync.source] = this.getSource(sync.source).pipe(
+            take(1),
+            concatMap((source) => {
+              if (source.hasServerSideAuth) {
+                // TODO: we cannot redirect to several third party login pages at the same time
```
mpellerin42, line 184: I don't understand the meaning of this todo, it seems like you're describing something you cannot do instead of what we should try to do 🤔.

PR #77, opened 2022-06-16
Nua key activity log page

mpellerin42:
- NUA Activity logs page
- Access account-nua from its own route
- Add activity button to understanding API key
- Use $gray-contrast instead of #707070


apps/app/src/app/account/account-nua/account-nua.component.ts
```diff
@@ -16,57 +15,68 @@ import { TokenDialogComponent } from '../../components/token-dialog/token-dialog
...
+        switchMap(() => this.nua.renewClient(client.client_id)),
       )
       .subscribe(({ token }) => {
         this.showToken(token);
         this.nua.updateClients();
       });
   }
 
+  goToActivity(client: NUAClient): void {
+    const currentPath = location.pathname;
```
ebrehault, line 55: This approach is a bit hacky, we have a navigation service to provide proper route path. Go that way:
```ts
accountUrl = this.stateService.account.pipe(
    filter((account) => !!account),
    map((account) => this.navigation.getAccountUrl(account!.slug)))
```

apps/app/src/app/account/account-nua/nua-activity/nua-activity.component.ts
```diff
@@ -0,0 +1,43 @@
...
+export interface Activity {
+  file: string;
+  timestamp: string;
+  actor: string;
+}
+
+@Component({
+  selector: 'app-nua-activity',
+  templateUrl: './nua-activity.component.html',
+  styleUrls: ['./nua-activity.component.scss'],
```
ebrehault, line 18: OnPush

ebrehault (changes requested):

apps/app/src/app/account/account-nua/account-nua.component.html
```diff
@@ -32,6 +32,12 @@
                              stfTooltip
                              stfTooltipText="account.generate_key">
           </app-button-action>
+          <app-button-action (click)="goToActivity(client)"
```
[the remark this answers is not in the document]
mpellerin42: Good idea!

apps/app/src/app/account/account-nua/account-nua.component.ts
```diff
@@ -56,8 +59,11 @@ export class AccountNUAComponent {
   }
 
   goToActivity(client: NUAClient): void {
-    this.stateService.account
+    this.isNuaActivityEnabled
```
ebrehault, line 62: a bit overkill (the button is invisible if the feature is not enabled), but no big deal

ebrehault (approved):

apps/app/src/app/account/account-nua/account-nua.component.ts
mpellerin42: I know, it's a security habit I have from Onna 😅 
And actually, I didn't go through it all as I didn't protect the ability to route to nua activity page with a guard, but I felt like that was good enough 😉

operramon (approved):

PR #78, opened 2022-06-17
Ericbrehault/sc 1571/implement the new navigation for account

ebrehault:


apps/app/src/app/account/account.component.scss
```diff
@@ -0,0 +1,8 @@
+:host > ::ng-deep app-account-kbs {
+  margin: 22px;
```
mpellerin42, line 2: should be rythm(2) I guess (which is 24px in base12)

mpellerin42 (approved):

apps/app/src/app/account/account.component.scss
mpellerin42: I saw some other pixel values everywhere in the code. It's probably not the most urgent thing to fix, specially if we go for a base 8 soonish, so I didn't commented the others…

operramon (approved):

PR #81, opened 2022-06-17
Oriolperramon8396/sc 1570/implement the new account dashboard

operramon:


apps/app/src/app/account/account-home/account-home.component.html
```diff
@@ -20,16 +20,42 @@ <h2>{{ 'account.characters' | translate }}</h2>
...
-        </a>
-      </div>
-
       <div class="container">
         <span class="totals"
               [innerHtml]="'account.total' | translate:{ total: (kbsTotal | async) || 0, public: (kbsPublic | async) || 0}"></span>
       </div>
+      <div class="container">
+        <h2>{{ 'account.total_queries' | translate }}</h2>
+        <div class="total-queries">{{ ((totalQueries | async) || '').toLocaleString() }}</div>
```
mpellerin42, line 30: Why not using Angular's builtin pipe for formatting numbers instead? They are listed in here: https://angular.io/api/common#pipes, I think the one you're looking for is DecimalPipe: https://angular.io/api/common/DecimalPipe

ebrehault (approved):

apps/app/src/app/components/charts/bar-chart/bar-chart.component.html
```diff
@@ -1,3 +1,8 @@
 <div class="bar-chart"
      #container
-     [id]="id"></div>
\ No newline at end of file
+     [id]="id">
+  <div class="tooltip"
+       [style.opacity]="showTooltip ? '1' : '0'"
+       [style.bottom.px]="tooltipBottom"
+       [style.left.px]="tooltipLeft">{{ (tooltipContent || '').toLocaleString() }}</div>
```
mpellerin42, line 7: same here: tooltip content is a number, so better to use DecimalPipe from angular

apps/app/src/app/account/account-home/account-home.component.html
operramon: looks good, I'll fix it the next PR

PR #87, opened 2022-06-21
gdrive oauth in electron

ebrehault:


apps/app/src/app/redirect/redirect.component.ts
```diff
@@ -1,18 +1,67 @@
 import { Component } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
 import { SDKService } from '@flaps/auth';
-import { take } from 'rxjs';
+import { injectScript } from '@flaps/core';
+import { take, tap, concatMapTo } from 'rxjs';
 
+declare var gapi: any;
```
mpellerin42, line 7: var?

ebrehault: it is a global variable created by the Google JS client which is injected dynamically so it is not like it is part of our nodes packages, it cannot be resolved by typescript or anything.
If we want typescript to be happy with it, we have to declare it as a global variable and `declare var` is the syntax we are supposed to used (you will also notice the `// <reference path="` things at the beginning which allows to load typings for these dynamically loaded libraries)

mpellerin42: thanks for the explanations :)

mpellerin42 (approved):

apps/app/src/app/redirect/redirect.component.ts
```diff
@@ -1,18 +1,67 @@
...
+
+const SCOPES = 'https://www.googleapis.com/auth/drive.metadata.readonly https://www.googleapis.com/auth/drive.readonly';
+const DISCOVERY_DOCS = ['https://www.googleapis.com/discovery/v1/apis/drive/v3/rest'];
 @Component({
   template: 'Redirecting to Nuclia Desktop…',
 })
 export class RedirectComponent {
   constructor(private sdk: SDKService, private route: ActivatedRoute) {
     this.route.queryParams.pipe(take(1)).subscribe((params) => {
       let redirectUrl: string = params['redirect'] || '';
```
operramon, line 18: what do you think about exit if param['redirect'] is different than 'nuclia-desktop://' or 'localhost' ?
Maybe someone could create malicious urls like 'nuclia.cloud/redirect?redirect=maliciuos.com'

ebrehault: waouh that's a very very good point!!!

ebrehault: fixed

PR #95, opened 2022-06-27
Show alert when entering an empty knowledge box

operramon:


apps/app/src/app/services/app.service.ts
```diff
@@ -40,12 +40,30 @@ export class AppService {
...
-      this.currentLocaleSubject.next(this.getLocaleFromLang(event.lang))
+      this.currentLocaleSubject.next(this.getLocaleFromLang(event.lang));
     });
   }
 
   private getLocaleFromLang(lang: string) {
     return this.config.getLocales().includes(lang) ? lang : 'en-US';
   }
 
+  checkEmptyKBAlert() {
```
mpellerin42, line 51: It took me a while to understand why you were mapping the counter to a boolean, so maybe renaming (and typing) this method would be better. 
I'm not sure what would be a good name for it though. `hasResources` ? `isEmptyKb`? `isKbStillEmptyAfterFirstDay`?

ebrehault (approved):

mpellerin42 (approved):

PR #97, opened 2022-06-28
[sc-1648]: allow to change the kb slug

mpellerin42:
Add readonly UID field, and editable slug field to kb profile page


apps/app/src/app/knowledge-box/knowledge-box-profile/knowledge-box-profile.component.ts
```diff
@@ -52,20 +54,23 @@ export class KnowledgeBoxProfileComponent implements OnInit, OnDestroy {
...
       description: this.kbForm.value.description,
-      // anonymization: this.useAnonymization ? 'anonymization' : '',
+      slug: newSlug,
     };
     (this.kb as WritableKnowledgeBox)
       .modify(data)
-      .pipe(concatMap(() => this.sdk.nuclia.db.getKnowledgeBox(this.account!.slug, this.kb!.slug!)))
+      .pipe(concatMap(() => this.sdk.nuclia.db.getKnowledgeBox(this.account!.slug, newSlug)))
       .subscribe((kb) => {
         this.stateService.setStash(kb);
```
ebrehault, line 75: don't we need to re-route? (as the current route contain the old slug, it can break navigation, no?)

libs/common/src/assets/i18n/ca.json
```diff
@@ -391,7 +393,7 @@
 	"stash.home.tags.title": "Gràfica de distribució de tags",
 	"stash.home.tags.total": "tags en total",
 	"stash.invited_user": "L'usuari {{user}} ha rebut un correu amb una invitació a participar a la caixa de coneixement",
-	"stash.kb_name_invalid": "El nom de la caixa de coneixement no és vàlid",
```
operramon, line 394: I see this message is used in other parts of the app to validate the kb name field.
What do you think about adding another text like "stash.kb_slug_invalid" ?

apps/app/src/app/knowledge-box/knowledge-box-profile/knowledge-box-profile.component.ts
```diff
@@ -17,12 +18,14 @@ export class KnowledgeBoxProfileComponent implements OnInit, OnDestroy {
   account: Account | undefined;
 
   kbForm = this.formBuilder.group({
-    title: ['', [Sluggable()]],
+    uid: [''],
+    slug: ['', [Sluggable()]],
```
operramon, line 22: I realize that "sluggable" validator is kind of confusing. It doesn't validate if a slug is valid,
It validates if a Kb name can be automatically converted into a valid Kb slug.

apps/app/src/app/knowledge-box/knowledge-box-profile/knowledge-box-profile.component.ts
mpellerin42: The current route contain the old slug yes, but as soon as you navigate somewhere else the slug is automatically updated without requiring re-routing.

ebrehault (approved):

libs/common/src/assets/i18n/ca.json
mpellerin42: Good catch!
That being said, the invalidity of the name of the KB was always done based on the fact the name would become a slug right? shouldn't we remove this validation and display another field showing the generated slug ?
I feel like this is something to be reviewed with Anna as well though. So for now I'll just put back `stash.kb_name_invalid` and add a new one for slug validation.

apps/app/src/app/knowledge-box/knowledge-box-profile/knowledge-box-profile.component.ts
mpellerin42: During my tests, I noticed the validator wasn't working very well indeed, or at least it didn't feel like it was working well. 
The thing is, our method `STFUtils.generateSlug` is transforming any text into a valid slug, and once saved the form is updated with the generated slug. So the current validator is good enough as it prevents the user to only enter invalid characters. For example, typing `#` is invalid because it would become an empty slug, but typing `#k` is valid because it will generate a slug `k` which is a valid one.

libs/common/src/assets/i18n/ca.json
operramon: I don't remember why we did it this way (maybe easy of use), Anna knows it better than me.

operramon (approved):

PR #103, opened 2022-06-29
webhook field when creating a NUA key

ebrehault:


apps/app/src/app/account/account-nua/client-dialog/client-dialog.component.html
```diff
@@ -11,9 +11,9 @@ <h1>{{ (editMode ? 'account.edit_nua_key' : 'account.create_nua_key') | translat
          fxLayout="row">
       <label for="client-title"
              fxFlex="24%">
-        {{ 'account.nua_client' | translate }}
+        {{ 'account.nua_client' | translate }}*
```
mpellerin42, line 14: is the star meaning mandatory field?

ebrehault: yes

PR #105, opened 2022-06-30
Nx projects workspace

mpellerin42:


mpellerin42:
Closing until we use same @nrwl package for all deps

PR #113, opened 2022-07-04
Replace auth lib by core lib

mpellerin42:
- Add index.ts in all folders of core lib
- Move all auth lib content to core lib
- Replace all @flaps/auth imports by @flaps/core
- Delete lib @flaps/auth
- Rename APIService to DeprecatedApiService


ebrehault (changes requested):
the `auth` lib is still mentioned in `.github/workflows/deploy.yml` line 54

mpellerin42:
> the `auth` lib is still mentioned in `.github/workflows/deploy.yml` line 54

good catch! I checked more broadly and also found and remove a mention of it in tsconfig

ebrehault (approved):

PR #117, opened 2022-07-06
Matpellerin/sc 1805/sistema toasts

mpellerin42:
- Sistema toasts
- Use Sistema toasts in dashboard app
- Improve sistema-demo welcome page


ebrehault (commented):
I think `SisToastService` is a good name but the file itself could be named just `toast.service.ts`.
`sis-toast.service.ts` sounds a bit too much considering it is in a library named `sistema`, that's unneeded noise, no?

mpellerin42:
> I think `SisToastService` is a good name but the file itself could be named just `toast.service.ts`. `sis-toast.service.ts` sounds a bit too much considering it is in a library named `sistema`, that's unneeded noise, no?

I'm using the cli to create new files, if we want the class to be named `SisToastService`, we need to run `nx g service sis-toast --project=sistema`

ebrehault (approved):

PR #118, opened 2022-07-06
extract paragraph texts

ebrehault:


libs/sdk-core/protobuf/README.md
```diff
@@ -2,7 +2,11 @@
 
 JavaScript utility to handle Nuclia protobuffer messages.
 
-`NucliaProtobufConverter` is a function accepting a buffer and returning a Promise that resolves to a parsed protobuffer message.
+## Functions
+
+- `NucliaProtobufConverter` is a function accepting a buffer and returning a Promise that resolves to a parsed protobuffer message.
+
+- `extractParagraphTexts` is function adding to each Paragraph object a `text` attribute containing the actual text of the paragraph (by default, Paragraph objects only contain the position of their first and last characters).
```
mpellerin42, line 9:
```suggestion
- `extractParagraphTexts` is a function adding to each Paragraph object a `text` attribute containing the actual text of the paragraph (by default, Paragraph objects only contain the position of their first and last characters).
```

libs/sdk-core/protobuf/index.mjs
```diff
@@ -24,3 +24,26 @@ export const NucliaProtobufConverter = (buffer) =>
...
+    ...field,
+    metadata: field.metadata
+      ? {
+          ...field.metadata,
+          metadata: field.metadata.metadata
+            ? {
+                ...field.metadata.metadata,
+                paragraphs: field.metadata.metadata.paragraphs?.map((paragraph) => ({
+                  ...paragraph,
+                  text: payload.extractedText[index]?.body?.text?.slice(paragraph.start || 0, paragraph.end).trim(),
```
mpellerin42, line 42: what happens if there is no `extractedText`? should you return an empty string in that case?

mpellerin42 (approved):

PR #123, opened 2022-07-07
Oriolperramon8396/sc 1803/allow to upload a local folder

operramon:


apps/desktop/src/app/sync/sources/folder.ts
```diff
@@ -0,0 +1,86 @@
...
+      nextPage: hasMoreResults ? this._getFiles(query, pageSize, offset + pageSize) : undefined,
+    });
+  }
+
+  private searchFiles(query: string) {
+    const regex = new RegExp(`(${query})`, 'i');
+    return this.files.filter((file) => regex.test(file.name));
+  }
+
+  download(resource: SyncItem): Observable<Blob> {
```
mpellerin42, line 79: based on your code, the typing should be `Observable<Blob | void>`

mpellerin42 (commented):
I don't think you need to add the folder icon in desktop app: this icon is included in sistema, so you can use: `<pa-icon name="folder">` to display it

apps/desktop/src/app/sync/sources/folder.ts
operramon: download is fixed

PR #124, opened 2022-07-08
[sc-1802] Desktop: Algolia integration

mpellerin42:


apps/desktop/src/app/sync/protobuf.ts
```diff
@@ -21,6 +21,21 @@ export const NucliaProtobufConverter = (buffer: Uint8Array) =>
     }
   });
 
+const BASE64_MARKER = ';base64,';
+export const convertDataURIToBinary = (dataURI: string) => {
+  const base64MarkerIndex = dataURI.indexOf(BASE64_MARKER);
+  const base64Index = base64MarkerIndex === -1 ? 0 : dataURI.indexOf(BASE64_MARKER) + BASE64_MARKER.length;
+  const base64 = dataURI.substring(base64Index);
```
ebrehault, line 28: that's really complex and difficult to understand
what about:
```ts
const base64 = dataURI.includes(BASE64_MARKER) ? dataURI.split(dataURI)[1] : dataURI;
```
?

apps/desktop/src/app/sync/protobuf.ts
```diff
@@ -21,6 +21,21 @@ export const NucliaProtobufConverter = (buffer: Uint8Array) =>
...
+  const raw = window.atob(base64);
+  const rawLength = raw.length;
+  const array = new Uint8Array(new ArrayBuffer(rawLength));
+
+  for (let i = 0; i < rawLength; i++) {
+    array[i] = raw.charCodeAt(i);
+  }
+  return array;
+};
+
```
ebrehault, line 38: I think it could be done with:
```ts
const array = Uint8Array.from(raw);
```

apps/desktop/src/app/sync/sync.service.ts
```diff
@@ -204,7 +205,36 @@ export class SyncService {
...
+        map((data: any) => {
+          const sync = this._queue.find((sync) => sync.fileUUIDs.includes(data.uuid));
+          let item: SyncItem | undefined;
+          if (sync) {
+            item = sync.files.find((file) => file.uuid === data.uuid);
+          }
+          return { data, sync, item } as { data: any; sync: Sync | null; item?: SyncItem };
+        }),
+        filter(({ sync, item }) => !!sync && !!item),
+        map(({ data, sync, item }) => ({ data, sync, item } as { data: any; sync: Sync; item: SyncItem })),
```
ebrehault: if you reverse these 3 steps, you do not need `if` anymore:
```ts
map((data: any) => ({ data, sync: this._queue.find((sync) => sync.fileUUIDs.includes(data.uuid)) })),
filter(({data, sync}) => !!sync),
map(({data, sync}) => ({ data, sync, item: sync.files.find((file) => file.uuid === data.uuid) })),
filter(({data, sync, item}) => !!item),
```
not a huge improvement though, more a matter of taste

apps/desktop/src/app/sync/protobuf.ts
mpellerin42: way better! thanks

apps/desktop/src/app/sync/protobuf.ts
mpellerin42: Sadly no, it's not that simple
image

ebrehault (approved):

PR #125, opened 2022-07-08
add affinity to deployments

ferpizza:


charts/manager/values.yaml
```diff
@@ -1,16 +1,21 @@
...
 hash: HASH_VERSION_TO_REPLACE
-resources:
-  limits:
-    memory: '50Mi'
-    cpu: '50m'
+
+resources: {}
+  # limits:
+  #   memory: '50Mi'
+  #   cpu: '50m'
```
mpellerin42, line 14: any reason to keep the commented code?

charts/app/values.yaml
```diff
@@ -1,17 +1,21 @@
...
 hash: HASH_VERSION_TO_REPLACE
-resources:
-  limits:
-    memory: '50Mi'
-    cpu: '50m'
+
+resources: {}
+  # limits:
+  #   memory: '50Mi'
+  #   cpu: '50m'
```
mpellerin42, line 14: commented code?

charts/manager/values.yaml
ferpizza: yes.
this file has the default values for the Helm chart. It's better to avoid fixing resources limits/allocation by means of default values, to force the user to choose their own (it's a widespread practice)

charts/app/values.yaml
ferpizza: see prior comment

charts/manager/values.yaml
mpellerin42: I meant: shouldn't we remove the code intead of keeping it commented?

PR #133, opened 2022-07-13
fix backend url when running prod desktop

ebrehault:


libs/core/src/lib/config/app.init.service.ts
```diff
@@ -70,9 +70,10 @@ export class AppInitService {
               });
             });
           }
-          const apiOrigin = config.production
-            ? location.origin.replace('manage.', '').replace('auth.', '')
-            : config.backend.apiOrigin;
+          const apiOrigin =
+            config.production && location.origin.startsWith('http')
```
operramon, line 74: would it be too much redundant to add a comment here?
Without context, I'm not sure if I would understand it's a condition for desktop app.

operramon (approved):

PR #146, opened 2022-07-25
Ericbrehault/sc 1925/allow to select all files from source

ebrehault:


apps/desktop/src/app/upload/select-files/select-files.component.ts
```diff
@@ -128,4 +130,35 @@ export class SelectFilesComponent implements AfterViewInit, OnDestroy {
...
+      });
+    } else {
+      this.selection = new SelectionModel(true, []);
+      this.selectionChange.emit(this.selection);
+      this.cdr?.markForCheck();
+    }
+  }
+
+  getAllFiles(): Observable<SyncItem[]> {
+    return (this.source as ISourceConnector).getFiles().pipe(
```
operramon, line 153: It selects the first two pages of results. Should it get more ? or two is enough ?

operramon (approved):

PR #148, opened 2022-07-26
Display labels on paragraphs

operramon:


libs/search-widget/src/viewer/paragraphs/Paragraphs.svelte
```diff
@@ -17,6 +19,10 @@
       paragraph: paragraph.paragraph,
     });
   };
+  const showLabels = nucliaState().widget.pipe(
```
ebrehault, line 22: I think we will always show labels (no matter if edition is enabled or not)

ebrehault (approved):

libs/search-widget/src/viewer/paragraphs/Paragraphs.svelte
operramon: ok, now they are always shown

PR #150, opened 2022-07-27
Allow to set labels on paragraphs

operramon:
Feature to set labels on paragraphs


ebrehault (approved):

libs/search-widget/src/components/label/Label.svelte
```diff
@@ -0,0 +1,52 @@
...
+  {#if removable}
+    <img on:click={remove} src={`${getCDN()}icons/circle-cross.svg`} alt="delete">
+  {/if}
+</div>
+
+<style>
+  .label {
+    display: inline-flex;
+    padding: 0 0.5em;
+    font-size: 0.75em;
```
mpellerin42, line 32: out of curiosity, why are we using `em` instead of `rem` on svelte components?

libs/search-widget/src/components/label/Label.svelte
```diff
@@ -0,0 +1,52 @@
...
+    flex: 1 1 0;
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+    filter: saturate(100) brightness(80%);
+  }
+  img {
+    flex: 0 0 auto;
+    margin-left: 0.5em;
+    width: 14px;
```
mpellerin42, line 49: shouldn't we be base 8 on svelte components as well? even though we can't use the rhythm map we have in angular, maybe we can stick with the corresponding pixels values (2, 4, 8, 12, 16,…)?

libs/search-widget/src/viewer/LabelMenu.svelte
```diff
@@ -0,0 +1,178 @@
...
+  const dispatch = createEventDispatcher();
+  export let position: { top: number, left: number } | undefined = undefined;
+  export let labels: Classification[] = [];
+
+  const savingLabels = viewerStore.savingLabels;
+  let selected: { [key: string]: boolean } = {};
+  $: selected = labels.reduce((acc, current) => {
+        acc[`${current.labelset}-${current.label}`] = true;
+        return acc;
+      }, {} as { [key: string]: boolean });
```
mpellerin42, line 21: weird indentation

mpellerin42 (approved):

libs/search-widget/src/viewer/LabelMenu.svelte
operramon: Fixed

PR #155, opened 2022-08-01
[sc-1887] Rename app into dashboard in code and deployment

mpellerin42:


ebrehault (approved):
We probably also want to rename `app` in `charts/app` and in the `frontend_deploy` repository (and all its files), but I recommend we do that when Fer is back.
That's not a blocker for this PR though.
(I will create another ticket for the chart change)

PR #161, opened 2022-08-04
[sc-2011] Edit resource improvements

mpellerin42:
- Improve hint component using Pastanaga + optional inverted color
- Improve edit resource templates using Pastanaga
- Fix spacing by removing duplicated scss function from stf
- Improve Label field and resource profile page


libs/common/src/assets/i18n/es.json
```diff
@@ -293,6 +293,7 @@
 	"reset.password_reset": "Se ha restablecido la contraseña",
 	"reset.title": "Crea contraseña",
 	"resource.access_source_file": "Acceder al documento de origen",
+	"resource.authorship": "Paternidad literaria",
```
operramon, line 296: What about change to "Autoría" ?

operramon (approved):

libs/common/src/assets/i18n/es.json
mpellerin42: no problem with me, I just took what babeledit was suggesting (it uses google translate)

PR #163, opened 2022-08-05
Fix [sc-1842]: rename widget

mpellerin42:
Renaming a widget is not allowed byt the backend, so we have to delete and re-create it with the new id.

### DIFF
--- a/apps/dashboard/src/app/widgets/edit/edit-widget.component.scss
+++ b/apps/dashboard/src/app/widgets/edit/edit-widget.component.scss
@@ -1,4 +1,4 @@
-@import 'apps/dashboard/src/variables';
+@import 'variables';
 
 .edit-widget {
   width: 770px;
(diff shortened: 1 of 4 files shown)


apps/dashboard/src/app/widgets/edit/edit-widget.component.scss
```diff
@@ -1,4 +1,4 @@
-@import 'variables';
+@import 'apps/dashboard/src/variables';
```
operramon, line 1: just to know, is it best to use absolute path ?

operramon (approved):

apps/dashboard/src/app/widgets/edit/edit-widget.component.scss
mpellerin42: oh good catch, I didn't notice that change which occurred automatically when I moved the files in the folder. I'll revert that as we want to keep `@import 'variables';`.

PR #172, opened 2022-08-09
[sc-2088] and [sc-2029] desktop improvements

mpellerin42:
- Fix [sc-2088]: disable upload for kb without write permission
- [sc-2029]: 
  - Fix upload form submission
  - Fix Desktop application name: missing space
  - Add switch account option in user menu
  - Replace folder image and update the connectors to look like cards
  - Add a rocket to the upload steps


ebrehault (approved):

apps/desktop/src/app/upload/upload.component.spec.ts
```diff
@@ -38,8 +45,11 @@ describe('UploadComponent', () => {
         ReactiveFormsModule,
         MatDialogModule,
         ConfirmFilesModule,
-        PaButtonModule,
-        PaTextFieldModule,
+        MockModule(PaButtonModule),
```
operramon, line 48: Just to know, is `MockModule` required for pastanaga modules ?

operramon (approved):

apps/desktop/src/app/upload/upload.component.spec.ts
mpellerin42: no it's not required. It's just a way to make test more performant by not loading everything for real. That being said, I didn't check the performances before/after using MockModule here, I just added them by habit.

operramon: Thanks for the info

PR #177, opened 2022-08-10
[sc-1837] Use Pastanaga in select account and kb pages

mpellerin42:
* Make a standalone component for back-button
* Display all kbs for managers (disabled when not a member)
* Fix empty box behavior when canceling adding a kb
* Replace STF by Pastanaga in select kb components

### DIFF
--- a/apps/dashboard/src/app/select/select.component.html
+++ b/apps/dashboard/src/app/select/select.component.html
@@ -1,38 +1,28 @@
 <div class="select"
      *ngIf="accounts && kbs">
-  <div class="select-container"
-       fxLayout="column"
-       fxFlex="0 0 auto">
+  <div class="select-container">
     <div class="logo">
-      <img src="assets/logos/logo_text.svg" />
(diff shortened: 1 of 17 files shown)


ebrehault (approved):

apps/dashboard/src/app/select/select.component.html
```diff
@@ -1,41 +1,31 @@
...
     <div class="select-content"
-         fxFlex="1 1 0"
          [@slideInOut]="selectKb">
       <div class="select-account"
-           *ngIf="!selectKb"
-           fxLayout="column">
-        <div class="select-account-list"
-             fxFlex="1 1 0">
+           *ngIf="!selectKb">
+        <div class="select-account-list">
```
operramon, line 11: I used to use long class names, feel free to change it if you think they're to long.

mpellerin42: No worries, I like descriptive names more than short names

PR #181, opened 2022-08-12
[sc-1837] use sistema icons in dashboard app

mpellerin42:
- Add some values to sistema rhythm tokens
- [sc-1837] Replace all icons in
  - entities page
  - filter-ontology (blindly as it seems fitlers are not accessible from the dashboard currently)
  - upload-bar
- Replace `app-button-actions` by pa-button and remove `app-button-actions` and its module from the dashboard


apps/dashboard/src/app/components/search-filters/search-filters.module.ts
```diff
@@ -9,6 +9,7 @@ import { STFExpanderModule } from '@flaps/pastanaga';
 import { SearchFiltersComponent } from './search-filters/search-filters.component';
 import { ButtonClearComponent } from './button-clear/button-clear.component';
 import { FilterOntologyComponent } from './filter-ontology/filter-ontology.component';
+import { PaIconModule } from '@guillotinaweb/pastanaga-angular';
 
 @NgModule({
```
operramon, line 14: I think SearchFilterModule is not used anymore, it corresponds to an old design. Maybe it can be deleted at some point.

operramon (approved):

apps/dashboard/src/app/components/search-filters/search-filters.module.ts
mpellerin42: Do you think I can delete it in this PR? I don't like keeping dead code in a repo

operramon: yes

mpellerin42: cool, done. Thanks

PR #183, opened 2022-08-12
[sc-2108] use pa icon in common lib

mpellerin42:
- Support Pastanaga style in Manager
- Use pa-icon in common lib
- Replace simple-select by dropdown-button in resource list
  - Create new dropdown-button standalone component in sistema
  - use dropdown-button in label-field, kb-process and resource list
  - Remove STF-simple-select component and module
- Remove STF tag component from component lib


apps/dashboard/src/app/resources/resource-list.component.ts
```diff
@@ -69,6 +74,8 @@ export class ResourceListComponent implements OnInit, OnDestroy {
   unsubscribeAll = new Subject<void>();
   refreshing = true;
 
+  ////////////////////////////////////////////////////////////////////////////////////////////////////////
+  // FIXME: those options are not used anywhere, can we remove them or is there a reason to keep them here?
```
operramon, line 78: I think the resource list page is supposed to have a `status` and `type` filters, but haven't been implemented in the back yet. I guess it can be deleted.

operramon (approved):

apps/dashboard/src/app/resources/resource-list.component.ts
mpellerin42: ok, let's keep them for now, I'll ask Ramon next week (or whenever they'll have finished to fix what they're trying to fix lately) about those filters and we'll remove them depending on the answer.

PR #185, opened 2022-08-16
[sc-2039] Chrome extension to push links to Nuclia

operramon:
Initial minimal version of Chrome extension to push links to Nuclia


libs/chrome-tools/manifest.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "Nuclia tools",
+  "description": "Upload web content to Nuclia search engine",
+  "version": "0.1",
```
mpellerin42: shouldn't it be semantic versioning with `major.minor.patch` format like `0.0.1`?

libs/chrome-tools/background.js
```diff
@@ -0,0 +1,37 @@
...
+  chrome.contextMenus.create({
+    id: 'NucliaContextMenu',
+    title: 'Upload to Nuclia',
+    targetUrlPatterns: ['https://*/*'],
+    contexts: ['link'],
+  });
+});
+
+chrome.contextMenus.onClicked.addListener((info) => {
+  chrome.storage.local.get(['kb', 'key'], ({ kb, key }) => {
```
mpellerin42: To prevent any conflict in the storage, we should maybe prefix the storage keys by `NUCLIA_` or something, don't you think?

libs/chrome-tools/manifest.json
operramon: Fixed

libs/chrome-tools/background.js
operramon: Fixed

mpellerin42 (approved):

PR #188, opened 2022-08-17
[sc-2105] sistema confirmation dialogs

mpellerin42:
- Adapt Pastanaga confirmation dialog for Sistema
- Use confirmation dialog in dashboard
- Fix kb switch dropdown width
- Temporary excluding sistema lib from the tests, the time for nx to fix https://github.com/nrwl/nx/issues/10537 by supporting jest 28 (see https://github.com/nrwl/nx/pull/10857)


apps/dashboard/src/app/account/account-users/account-users.component.ts
```diff
@@ -83,50 +80,41 @@ export class AccountUsersComponent implements OnDestroy {
...
-      };
-      const dialogRef = this.dialog.open(STFConfirmComponent, {
-        width: '420px',
-        data: data,
-      });
-      dialogRef
-        .afterClosed()
-        .pipe(
-          takeUntil(this.unsubscribeAll),
+      this.modalService
```
operramon, line 83: When using this modal an error of type "ExpressionChangedAfterItHasBeenCheckedError" appears in the console.

operramon (approved):

apps/dashboard/src/app/account/account-users/account-users.component.ts
mpellerin42: good catch! let me fix that

mpellerin42: ok, I'm not able to find where the problem is coming from right now, so I'll open a ticket to solve this. As you said, this is not super urgent to fix

PR #192, opened 2022-08-17
[sc-2111] Modify thumbnail dimensions in search results

operramon:
- Display the thumbnails without modifying its aspect ratio or cropping the image.
- Add a background color to the thumbnails container, so that all thumbnails look uniform.

### DIFF
--- a/libs/search-widget/src/widgets/row/Row.svelte
+++ b/libs/search-widget/src/widgets/row/Row.svelte
@@ -82,7 +82,7 @@
     </div>
     <div class="block-4">
       {#if thumbnail}
-        <div class="thumbnail">
+        <div class="thumbnail" class:semantic>
           <img src={thumbnail} alt="Thumbnail" />
         </div>
       {/if}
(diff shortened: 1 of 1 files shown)


libs/search-widget/src/widgets/row/Row.svelte
```diff
@@ -163,16 +163,22 @@
     position: relative;
     width: 100%;
     height: 0;
-    padding-top: 50%;
+    padding-top: 80%;
+    background-color: var(--color-dark-light);
+  }
+  .thumbnail.semantic {
+    background-color: #f0f0f0;
```
mpellerin42, line 170: this is `color-neutral-lighter` in our palette https://nuclia.github.io/frontend/palette
I know we use svelte for our widget, but maybe we could find a way to have our CSS variables corresponding to the same palette in both our widget and apps.
It's ok to fix that in another PR though, as we'll probably have to apply such a fix to a lot of places in the widget

mpellerin42 (approved):

libs/search-widget/src/widgets/row/Row.svelte
operramon: Sounds good, it make sense to use the same colors.

PR #195, opened 2022-08-18
[sc-2121] Create a thumbnail component

operramon:
- Create a thumbnail component to be used everywhere a thumbnail is displayed.

### DIFF
--- a/libs/search-widget/src/widgets/row/Thumbnail.svelte
+++ b/libs/search-widget/src/widgets/row/Thumbnail.svelte
@@ -3,7 +3,7 @@
   import { getFile } from '../../core/api';
 
   export let src: string;
-  export let ratio: number = 5/4;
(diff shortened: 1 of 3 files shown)


libs/search-widget/src/widgets/row/Thumbnail.svelte
```diff
@@ -0,0 +1,37 @@
...
+  let thumbnail: string;
+  if (src) {
+    getFile(src).subscribe((url) => (thumbnail = url));
+  }
+  onDestroy(() => {
+    if (thumbnail) URL.revokeObjectURL(thumbnail);
+  });
+</script>
+
+<div class="thumbnail" style:padding-top="{`${Math.round(1/ratio*100)}%`}">
```
mpellerin42, line 17: I'm a bit surprised by this padding-top: does it mean we have more grey background on the top than on the bottom?

mpellerin42 (approved):

PR #203, opened 2022-08-23
[sc-2084] Add spinner on KB creation dialog

mpellerin42:


operramon (approved):

libs/sistema/src/lib/progress/spinner/spinner.component.ts
```diff
@@ -0,0 +1,22 @@
...
+  set size(value: Size | undefined) {
+    if (value) {
+      this._size = value;
+    }
+  }
+  get size() {
+    return this._size;
+  }
+
+  private _size: Size = 'medium';
```
operramon, line 21: just out of curiosity, is it necessary to define "_size" ?
would `@Input() size: Size = 'medium'` be the same ?

mpellerin42: this is to make sure to always have a defined size. Using a setter, if by mistake a developer set `<nsi-spinner size="">` then the size of the spinner will still be medium.
It might be over precaution, it's just a habit I took after some errors I got in my past components usage and some review I got on the PRs I made at that time. Sometime it will be useful, and sometime not really needed but always doing the same is easier to avoid the mistakes when it's actually needed.

PR #208, opened 2022-08-26
Load CSS variables from the widgets

mpellerin42:
Don't duplicate the CSS variables in all widgets, load them on the fly instead. CSS variables will be stored in the CDN


libs/search-widget/src/Widget.svelte
```diff
@@ -68,7 +74,13 @@
...
     lang = lang || window.navigator.language.split('-')[0] || 'en';
     setLang(lang);
-    style = nucliaState().customStyle;
+
+    // Load CSS variables (must be done after the CDN was set) and custom styles
+    getCssVariablesAsText().pipe(
+      switchMap((cssVariables) => nucliaState().customStyle.pipe(
+        map(customStyles => `${cssVariables} ${customStyles}`)
+      ))
+    ).subscribe((css) => style = css);
```
ebrehault, line 83: Maybe we could have a function that would do getCssVariablesAsText + nucliaState().customStyle to return the resulting css.
That way we would not have to repeat this block 3 times.

ebrehault (approved):

libs/search-widget/src/Widget.svelte
mpellerin42: Yes, I hesitated on that as I didn't know if nucliaState() was available from the core… Let me fix this.

PR #209, opened 2022-08-26
[sc-2136] avoid unnecessary suggest calls

mpellerin42:
- Trigger suggestion between each word
- Trigger suggestion after 500ms of inactivity when trimmed query changed

### DIFF
--- a/libs/search-widget/src/widgets/search-input/SearchInput.svelte
+++ b/libs/search-widget/src/widgets/search-input/SearchInput.svelte
@@ -143,13 +143,13 @@
     padding: var(--search-bar-padding);
   }
 
-  .search-bar-widget:focus {
+  .search-bar-widget:focus,
+  .embedded-search:focus {
     border: var(--search-bar-border-focus);
   }
 
   .embedded-search {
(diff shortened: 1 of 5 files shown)


libs/search-widget/src/widgets/search-input/SearchInput.svelte
```diff
@@ -38,13 +38,15 @@
   });
 
   $: {
-    if (value.trim() !== previous.trim()) {
+    const decodedValue = decodeURIComponent(value);
```
operramon, line 41: Maybe it's not safe to use `decodeURIComponent` without a try catch.
Yesterday Eric fixed a similar bug related to it in #207

operramon (approved):

libs/search-widget/src/widgets/search-input/SearchInput.svelte
mpellerin42: You're right. Also Eric explained me the expected behavior of the suggestion so I'll rework the PR. This decode was actually not needed and will be removed in my next commit

ebrehault (approved):

PR #210, opened 2022-08-26
[sc-2133] Allow to select label from Chrome extension

operramon:
- Extend the context menu with options to select a label


libs/chrome-ext/src/lib/background.js
```diff
@@ -4,34 +4,110 @@ try {
...
+    .db.getKnowledgeBox()
+    .pipe(rxjs.switchMap((kb) => kb.createLinkResource({ uri: url }, { classifications: labels })))
+    .subscribe();
+}
+
+function getSDK(kb, key) {
+  return new NucliaSDK.Nuclia({
+    backend: 'https://nuclia.cloud/api',
     knowledgeBox: kb,
     zone: 'europe-1',
```
ebrehault, line 110: I missed that on your first PR, but the zone should be part of the settings, just like the kb id or the API key.
(Of course, you can set it to `europe-1` by default for now.)

ebrehault (approved):

libs/chrome-ext/src/lib/background.js
operramon: Ok, I added a zone field with a default value of `europe-1`

ebrehault (approved):

PR #216, opened 2022-08-31
allow to import youtube channels/playlists

ebrehault:


libs/chrome-ext/src/lib/background.js
```diff
@@ -104,6 +123,52 @@ function uploadLink(settings, url, labels) {
...
+    .then((videos) => selectVideos(videos, labels));
+}
+
+function selectVideos(videos, labels) {
+  chrome.storage.local.set({ videos });
+  chrome.tabs.create({ url: `youtube/selection.html?labels=${JSON.stringify(labels)}` });
+}
+
+function uploadLinksList(list, labels) {
+  if (!(list && list.length && list.length > 0)) {
```
mpellerin42, line 166: I find this way of doing the exclusion a bit hard to read, why not using `||` instead of the `&&` so you can make it clearer?
```suggestion
  if (!list || !list.length || list.length === 0) {
```

libs/chrome-ext/src/lib/options/options.html
```diff
@@ -1,34 +1,59 @@
...
+    <fieldset>
+      <legend>YouTube</legend>
+      <div class="field">
+        <label for="youTubeKey">YouTube API key:</label>
+        <input id="youTubeKey"
+               type="text"
+               autocomplete="off">
+        <div class="help">Enter a YouTube Data API</div>
+      </div>
+    </fieldset>
```
mpellerin42, line 47: is this new fieldset required for all use case? if not, should we display something explaining when you need it?

libs/chrome-ext/src/lib/style.css
```diff
@@ -6,6 +6,10 @@ p {
   margin: 0 0 2rem 0;
 }
 
+fieldset {
+  border-style: dashed;
```
mpellerin42, line 10: dashed?
I didn't see the final result, but don't you think a 1px border with a light grey color (like #f0f0f0 or #f7f7f8) would have a better look and feel?

libs/chrome-ext/src/lib/style.css
```diff
@@ -37,7 +41,11 @@ button {
   border-radius: 4px;
   cursor: pointer;
 }
-
+button.secondary {
+  border: 1px solid #ff0066;
+  color: #ff0066;
+  background-color: #fff;
```
mpellerin42, line 47: I know you don't have access to the CSS variables we have in Sistema, but I would still recommend you to choose the colors in our palette.
https://nuclia.github.io/frontend/palette

libs/chrome-ext/src/lib/style.css
```diff
@@ -49,3 +57,20 @@ button {
   vertical-align: middle;
   font-weight: bold;
 }
+
+header {
+  background-color: #0d0d0d;
+}
+header img {
+  height: 50px;
```
mpellerin42, line 65: try to stick to base 8, even without the rhythm. so 48px is better than 50px ;)

libs/chrome-ext/src/lib/style.css
```diff
@@ -49,3 +57,20 @@ button {
   vertical-align: middle;
   font-weight: bold;
 }
+
+header {
+  background-color: #0d0d0d;
```
mpellerin42, line 62: color palette

libs/chrome-ext/src/lib/options/options.html
```diff
@@ -37,12 +38,17 @@
     </fieldset>
     <fieldset>
       <legend>YouTube</legend>
+      <p>
+        Optional. When a <a href="https://developers.google.com/youtube/v3/getting-started">YouTube API key</a> is
+        defined, you can index all videos from a YouTube channel or YouTube.
```
mpellerin42, line 43: `from a YouTube channel or YouTube.` didn't you forget a word at the end? like `YouTube channel or YouTube play list.`?

libs/chrome-ext/src/lib/options/options.html
```diff
@@ -37,12 +38,17 @@
...
+      </p>
       <div class="field">
         <label for="youTubeKey">YouTube API key:</label>
         <input id="youTubeKey"
                type="text"
                autocomplete="off">
-        <div class="help">Enter a YouTube Data API</div>
+        <div class="help">Enter a YouTube Data API.
+          playlist.</div>
```
mpellerin42, line 51: ah, I think I found the missing word XD

mpellerin42 (approved):

PR #233, opened 2022-09-06
Ericbrehault/sc 2244/use web component in widget demo

ebrehault:


libs/search-widget/src/_video-widget/VideoTile.svelte
```diff
@@ -125,31 +137,27 @@
...
-     class:showAllResults
-     bind:this={tileElement}
-     style:max-height="{expandedHeight}"
+<svelte:window bind:innerWidth />
+<div
+  class="sw-video-tile"
+  class:expanded
+  class:showAllResults
+  bind:this={tileElement}
+  style:max-height={expandedHeight}
```
mpellerin42, line 146: I noticed this kind of reformatting in a lot of places, would be nice if there was a way to configure your formatter to always keep the first attribute on the same line as the tag…

apps/search-widget-demo/src/App.svelte
```diff
@@ -1,13 +1,18 @@
...
-  import { setLang } from 'libs/search-widget/src/core/i18n';
+  import { setLang } from '../../../libs/search-widget/src/core/i18n';
   import { onMount } from 'svelte';
   import { NucliaWidget } from '../../../libs/search-widget/src';
   import { NucliaSearchBar, NucliaSearchResults } from '../../../libs/search-widget/src/_video-widget';
 
-  let selected = 'two-widgets';
-  let widget: NucliaWidget;
+  let selected = 'input';
+  console.log(NucliaWidget, NucliaSearchBar, NucliaSearchResults);
```
mpellerin42, line 8: do we want to keep this console.log?

ebrehault: yes, we need to do something with these classes because if we don't, the compiler will ignore the import and it won't work (the compiler just thinks `<nuclia-search>` is a component that might or might not exist, it does not understand it is actually provided by the `NucliaSearch` class)

mpellerin42 (approved):

PR #236, opened 2022-09-07
[sc-2242] better css strategy

mpellerin42:
- Keep paragraphs centered when there is only a few of them
- Load CSS variables through SCSS

### DIFF
--- a/libs/search-widget/src/Widget.svelte
+++ b/libs/search-widget/src/Widget.svelte
@@ -76,7 +76,7 @@
     setLang(lang);
 
     loadFonts();
-    // Load CSS variables (must be done after the CDN was set) and custom styles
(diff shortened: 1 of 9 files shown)


libs/search-widget/src/Widget.svelte
```diff
@@ -78,7 +77,7 @@
 
     loadFonts();
     // Load CSS variables (must be done after the CDN was set) and custom styles
```
ebrehault, line 79: the comment is not accurate anymore, is it?

ebrehault (approved):

libs/search-widget/src/Widget.svelte
mpellerin42: good point

PR #244, opened 2022-09-12
display announce from github.com/nuclia/status

ebrehault:


apps/dashboard/src/app/app.component.ts
```diff
@@ -62,7 +63,8 @@ export class AppComponent implements AfterViewInit, OnInit, OnDestroy {
     if (this.config.getVersion()) {
       this.version = this.config.getVersion();
     }
-    this.checkStatusAlert();
+    this.checkMessages(true);
+    this.checkMessages(false);
```
mpellerin42, line 67: this is a bit weird and difficult to understand… why do we need to call the method twice with true and false? 
Could we have well named method doing these call instead so we can understand exactly what we're trying to achieve here?

apps/dashboard/src/app/components/messages/message-modal.component.html
```diff
@@ -0,0 +1,8 @@
+<pa-modal-dialog>
+  <pa-modal-title>{{ data?.title | translate }}</pa-modal-title>
+  <pa-modal-content [innerHtml]="data?.['message']"></pa-modal-content>
+  <pa-modal-footer>
+    <pa-button kind="primary"
+               (click)="modal.close()">{{ 'pastanaga.confirm' | translate}}</pa-button>
```
mpellerin42, line 6: On the previous implementation, the label was "ok". Does changing it to "Confirm" makes sense?

apps/dashboard/src/app/components/messages/message-modal.component.scss
```diff
@@ -0,0 +1,2 @@
+@import 'variables';
+@import '../../../../../../libs/pastanaga-angular/projects/pastanaga-angular/src/styles/theme/_modal.tokens.scss';
```
mpellerin42, line 2: why do you need this import? `_modal.tokens ` is not already imported in `variables`?

libs/core/src/lib/analytics/tracking.service.ts
```diff
@@ -90,15 +91,17 @@ export class STFTrackingService {
     return this.postHogService.getEnabledFeatures();
   }
 
-  getStatusAlert(): Observable<string> {
+  getStatusMessage(alert: boolean): Observable<string> {
```
mpellerin42, line 94: ok now I understand the `true`/`false` call. It would be easier to understand and easier to extend by using a type instead. like `'status' | 'announce'`.

mpellerin42 (changes requested):

mpellerin42 (approved):

PR #248, opened 2022-09-13
[sc-2184] Allow to authenticate to Nuclia from Chrome extension

operramon:
- Replace api key field with a login button
- Replace knowledge box Id field with an account and knowledge box selectors

### DIFF
--- a/apps/dashboard/src/app/redirect/redirect.component.ts
+++ b/apps/dashboard/src/app/redirect/redirect.component.ts
@@ -8,17 +8,25 @@ declare var gapi: any;
 declare var google: any;
 
 const AUTHORIZED_REDIRECTS = ['nuclia-desktop://', 'http://localhost:4200'];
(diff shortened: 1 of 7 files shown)


apps/dashboard/src/app/redirect/redirect.component.ts
```diff
@@ -8,17 +8,25 @@ declare var gapi: any;
 declare var google: any;
 
 const AUTHORIZED_REDIRECTS = ['nuclia-desktop://', 'http://localhost:4200'];
+const AUTHORIZED_REDIRECTS_REGEX = [/^chrome\-extension\:\/\/[a-z]+\/options\/options\.html$/];
```
mpellerin42, line 11: this name seems a bit off to me, doesn't seem to be an authorized redirects regexp, seems to be a from chrome extension regexp 🤔

ebrehault (approved):

apps/dashboard/src/app/redirect/redirect.component.ts
operramon: ok, at the moment I keep the name. If at some point we come up with a better name it can be changed.

PR #272, opened 2022-09-21
[sc-2429] Allow to trigger a search from outside the widget

operramon:


libs/search-widget/src/_video-widget/SearchBar.svelte
```diff
@@ -2,7 +2,7 @@
 
 <script lang="ts">
   import type { KBStates } from '@nuclia/core';
-  import { resetStore } from '../core/store';
+  import { resetStore, nucliaStore } from "../core/store";
```
ebrehault, line 5: prettier got weird, but no need to fix it now

ebrehault (approved):

libs/search-widget/src/_video-widget/SearchBar.svelte
operramon: I fix it now, otherwise I'll forget

PR #275, opened 2022-09-23
allow to set, add and delete entities

ebrehault:


operramon (approved):

libs/sdk-core/src/lib/db/resource.ts
```diff
@@ -155,6 +157,55 @@ export class Resource implements IResource {
...
 
+  setLabels(
+    fieldId: string,
+    fieldType: string,
+    paragraphId: string,
+    labels: Classification[],
+    preserve = true,
+  ): Observable<void> {
+    const fieldmetadata = preserve && this.fieldmetadata ? [...this.fieldmetadata] : [];
+    fieldmetadata.push({
```
mpellerin42, line 168: I have a doubt with this way of preserving the fieldmetadata: when preserve is true, if you call `setLabels` or `setEntities` several times, you will push a new fieldMetadata for labels instead of replacing the previous one as it was done before in `setLabels` from api.ts

mpellerin42 (approved):

mpellerin42 (approved):

PR #276, opened 2022-09-26
[sc-2409] entities annotation

mpellerin42:


ebrehault (approved):

libs/search-widget/src/viewer/Entities.svelte
```diff
@@ -26,47 +53,29 @@
...
+          <Icon name="chevron-left"/>
+        </div>
       </button>
       <ul>
-        {#each group[1] as entity}
-          <li tabIndex="0" role="button" on:click={() => search(entity)} on:keyup={(e) => onKeyUp(e, entity)}>
+        {#each group.entities as entity}
+          <li tabIndex="0" role="button"
+              class:clickable={!$annotationMode}
+              on:click={() => search(entity)}
```
operramon, line 77: A very small issue, click event is still triggered if $`annotationMode` is true

mpellerin42: good catch, thanks @operramon 
I noticed some bugs on the video widget as well, I fix this in another PR

PR #278, opened 2022-09-27
[sc-2449] Fix label annotation

operramon:
- Display two types of labels: labels from computed metadata and annotated labels from UserFieldMetadata field.
- Don't allow to modify computed labels. Only annotated labels can be modified.

### DIFF
--- a/libs/search-widget/src/viewer/viewer.store.ts
+++ b/libs/search-widget/src/viewer/viewer.store.ts
@@ -213,11 +213,12 @@ export function clearSearch() {
 
 export function setParagraphLabels(labels: Classification[], paragraph: WidgetParagraph) {
   const resource = viewerStore.resource.getValue();
+  const field = viewerStore.currentField.getValue();
   const saving = viewerStore.savingLabels.getValue();
-  if (!resource || saving) return;
+  if (!resource || !field || saving) return;
   viewerStore.savingLabels.next(true);
   const paragraphId = getParagraphId(resource.id, paragraph);
-  setLabels(resource, paragraph.fieldId, paragraph.fieldType.slice(0, -1), paragraphId, labels)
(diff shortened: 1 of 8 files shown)


libs/search-widget/src/viewer/viewer.store.ts
```diff
@@ -184,6 +211,25 @@ export function clearSearch() {
   viewerStore.onlySelected.next(false);
 }
 
+export function setParagraphLabels(labels: Classification[], paragraph: WidgetParagraph) {
+  const resource = viewerStore.resource.getValue();
+  const saving = viewerStore.savingLabels.getValue();
+  if (!resource || saving) return;
+  viewerStore.savingLabels.next(true);
+  const paragraphId = getParagraphId(resource.id, paragraph);
+  setLabels(resource, paragraph.fieldId, paragraph.fieldType.slice(0, -1), paragraphId, labels)
```
mpellerin42, line 220: `paragraph.fieldType.slice(0, -1)` I don't think this is correct as first letter of fieldType is not always the same as the one letter expected by the backend. For example `link` is `u`.
We know we have only one field type and field id for all paragraphs displayed on the viewer currently (because we don't display several fields at once in the viewer), so we stored the current field in the store for entities: `viewerStore.currentField`

ebrehault, line 220: you can get `fieldId` and `fieldType` from `viewerStore.currentField`

ebrehault (approved):

libs/search-widget/src/viewer/viewer.store.ts
operramon: Fixed

operramon: Fixed

mpellerin42 (approved):

PR #280, opened 2022-09-28
Ability to scroll in entity list

mpellerin42:


ebrehault (approved):

libs/search-widget/src/viewer/Metadata.svelte
```diff
@@ -34,6 +35,14 @@
     links = getLinks(resource);
   }
 
+  function setModalContentHeight() {
+    modalContentHeight = document.querySelector('.modal-content').getBoundingClientRect().height;
```
operramon, line 39: This line returns an error when using the widget in the dashboard, guess it's related to shadow DOM

mpellerin42: arf I'm going too quick without heavily testing things, my bad. Thanks for raising this @operramon , I'll rework it

PR #294, opened 2022-10-04
[sc-2554] Avoid upload of unsupported YouTube links

operramon:
An error message is displayed when the user tries to upload a Youtube link that is not a video, playlist or channel.

### DIFF
--- a/libs/chrome-ext/src/lib/utils.js
+++ b/libs/chrome-ext/src/lib/utils.js
@@ -1,0 +1,26 @@
+
+
+const isYoutubeUrl = (url) => {
+  return /^(?:https?:)?(?:\/\/)?(?:youtu\.be\/|(?:www\.|m\.)?youtube\.com\/)/.test(url);
+}
+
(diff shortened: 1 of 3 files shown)


libs/chrome-ext/src/lib/utils.js
```diff
@@ -0,0 +1,26 @@
+
+
+const isYoutubeUrl = (url) => {
+  return /^(?:https?:)?(?:\/\/)?(?:youtu\.be\/|(?:www\.|m\.)?youtube\.com\/)/.test(url);
```
mpellerin42, line 4: Here the method is creating the regexp everytime it is called. It's better to create the regexp as a const outside the method so the method always reuse the same instance of the regexp (you can apply this comment to all the methods using a regexp)

mpellerin42 (approved):

libs/chrome-ext/src/lib/utils.js
operramon: ups I didn't pay attention to it, now it's fixed.

PR #296, opened 2022-10-04
[sc-2530] Display success and error feedback in Chrome extension

operramon:


libs/chrome-ext/src/lib/background.js
```diff
@@ -207,7 +203,7 @@ function uploadLinksList(list, labels) {
   if (!list || !list.length || !list.length > 0) {
     return;
   }
-  getSettings().then((settings) => list.forEach((url) => uploadLink(settings, url, labels)));
+  getSettings().then((settings) => list.forEach((url) => uploadLink(settings, url, labels).subsceribe()));
```
ebrehault, line 206: typo `subsceribe`

ebrehault (changes requested):

libs/chrome-ext/src/lib/background.js
operramon: Fixed

PR #297, opened 2022-10-05
Ericbrehault/sc 2577/some paragraph annotations are not saved

ebrehault:


libs/search-widget/src/core/api.ts
```diff
@@ -127,6 +127,7 @@ export const setLabels = (
   if (!nucliaApi) {
     throw new Error('Nuclia API not initialized');
   }
+  console.log(paragraphId, labels);
```
operramon, line 130: A console log 🙂

operramon (approved):

PR #304, opened 2022-10-07
[sc-2581] Login infinite loop in desktop

operramon:


libs/core/src/lib/api/user.service.ts
```diff
@@ -29,6 +28,8 @@ export class UserService {
     return this.sdk.nuclia.db.getWelcome().pipe(
       catchError((error) => {
         if (error.status === 403 || error.status === 400) {
+          this.authService.setNextParams(this.route.snapshot.queryParams);
+          this.authService.setNextUrl(new URL(window.location.href).pathname);
```
mpellerin42, line 32: I don't really understand why we need these two lines (probably because I'm not familiar with the user flow you're fixing). Do you mind explaining me their purpose in this context?

mpellerin42 (approved):

ebrehault (approved):

PR #305, opened 2022-10-07
Support NucliaDB in standalone mode

ebrehault:


libs/sdk-core/src/lib/auth/auth.ts
```diff
@@ -26,7 +26,9 @@ export class Authentication implements IAuthentication {
   }
 
   getAuthHeaders(): { [key: string]: string } {
-    return this.nuclia.options.apiKey
+    return this.nuclia.options.standalone
+      ? { 'X-NUCLIADB-ROLES': 'READER' }
+      : this.nuclia.options.apiKey
       ? { 'X-STF-Serviceaccount': `Bearer ${this.nuclia.options.apiKey}` }
       : this.getToken()
       ? { Authorization: `Bearer ${this.getToken()}` }
```
mpellerin42, line 34: FYI I remember from my previous job this kind of imbrication of conditional ternary operator was raising ESlint error: https://eslint.org/docs/latest/rules/no-nested-ternary#rule-details
That being said, having it multiline like this makes it quite understandable to me here.

libs/search-widget/src/widgets/search-widget/Widget.svelte
```diff
@@ -136,21 +141,22 @@
...
-           --modal-height="var(--resource-modal-height)"
-           --modal-height-md="var(--resource-modal-height-md)"
+    <Modal
+      show={showModal}
+      on:close={closeModal}
+      closeButton={true}
+      --modal-width="var(--resource-modal-width)"
+      --modal-width-md="var(--resource-modal-width-md)"
+      --modal-height="var(--resource-modal-height)"
+      --modal-height-md="var(--resource-modal-height-md)"
```
mpellerin42, line 157: I really don't like this kind of reformatting. It seems like our IDE doesn't play using the same rules: seems like yours is forcing to have the attributes below the tag in svelte.

PR #311, opened 2022-10-13
[sc-2578] Disable entity annotation on safari

mpellerin42:
-  Display entity annotation button only when this feature is enabled
- Disable entity annotation on safari

### DIFF
--- a/libs/search-widget/src/old-components/viewer/Metadata.svelte
+++ b/libs/search-widget/src/old-components/viewer/Metadata.svelte
@@ -11,7 +11,7 @@
   import { onDestroy } from 'svelte';
   import {
     annotationMode,
-    annotations,
+    annotations, entityAnnotationFeatureEnabled,
     selectedFamily,
   } from '../../core/stores/annotation.store';
   import {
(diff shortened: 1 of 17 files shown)


libs/search-widget/src/widgets/search-widget/Widget.svelte
```diff
@@ -110,6 +111,7 @@
           activateEditLabelsFeature();
         }
       }),
+      canAnnotateEntities.subscribe(canAnnotate => entityAnnotationFeatureEnabled.set(canAnnotate))
```
operramon, line 114: one question, if canAnnotateEntities and entityAnnotationFeatureEnabled have the same value, do we need both to exist ?

operramon (approved):

libs/search-widget/src/old-components/viewer/Metadata.svelte
```diff
@@ -23,11 +23,14 @@
     summaries,
   } from '../../core/stores/resource.store';
   import { entityGroups } from '../../core/stores/entities.store';
-  import Modal from '../../common/modal/Modal.svelte';
+  import ConfirmDialog from "../../common/modal/ConfirmDialog.svelte";
 
   let entitiesBackup: string;
   let customEntitiesBackup;
 
+  const isSafari = navigator.userAgent.includes('Safari');
```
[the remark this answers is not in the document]
mpellerin42: good catch!

libs/search-widget/src/widgets/search-widget/Widget.svelte
mpellerin42: you're right, we need only one.

PR #317, opened 2022-10-17
[sc-2733] Labels classification column in resource list

mpellerin42:
- Label classification column
- Visible columns dropdown


libs/sdk-core/src/lib/db/resource.ts
```diff
@@ -29,6 +29,7 @@ import { search } from './search';
 export interface ReadableResource extends IResource {}
 export class ReadableResource implements IResource {
   data: ResourceData = {};
+  resourceLabels: SimpleLabel[] = [];
```
ebrehault, line 32: It looks more like something which helps in a specific use case (=our dashboard) than something people would need when using the SDK.

libs/sdk-core/src/lib/db/resource.models.ts
```diff
@@ -69,6 +69,10 @@ export interface LabelValue {
   label: string;
 }
 
+export interface SimpleLabel extends LabelValue {
+  color: string;
```
ebrehault, line 73: labels do not have color in the API, so this model does not belong here

libs/sdk-core/src/lib/db/resource.ts
mpellerin42: Right, I reworked my commit to avoid this

ebrehault (approved):

PR #318, opened 2022-10-17
[sc-2651] Connector for Amazon S3

operramon:
Add AWS S3 source connector to desktp app


apps/desktop/src/app/sync/models.ts
```diff
@@ -19,8 +19,9 @@ export interface ISourceConnector {
   hasServerSideAuth: boolean;
   resumable: boolean;
   getParameters(): Observable<Field[]>;
+  handleParameters?(params: ConnectorParameters): void;
```
mpellerin42, line 22: shouldn't the `?` be just before the `:` ?
```handleParameters(params: ConnectorParameters)?: void;```

File: apps/desktop/src/app/sync/sources/s3.ts [hunk 2]
```diff
@@ -0,0 +1,140 @@
...
+        type: 'text',
+        required: true,
+      },
+      {
+        id: 'secret_access_key',
+        label: 'Secret access key',
+        type: 'text',
+        required: true,
+      },
+    ]);
```
mpellerin42, line 61: should we translate the labels from this list?

apps/desktop/src/app/sync/sources/s3.ts
```diff
@@ -0,0 +1,140 @@
...
+  private filterResult(result: _Object, query?: string) {
+    if (query && result.Key) {
+      const regex = new RegExp(`(${query})`, 'i');
+      return regex.test(result.Key);
+    }
+    return true;
+  }
+
+  download(resource: SyncItem): Observable<Blob> {
+    const command = new GetObjectCommand({
```
ebrehault, line 132: Actually, the objective with Amazon S3 (and with Google Cloud Service) is not to download the physical and then upload it back to Nuclia like we did with DropBox or GDrive.
We want to use them as cloud file storage, so we want to use the Nuclia cloud file support, meaning we will create a FileField by providing its direct URL to Amazon S3 (and also some extra_headers to provide an auth token or whatever Amazon S3 in order to give access to the file)

apps/desktop/src/app/sync/models.ts
operramon: If `?` is at the end then Typescript complains, I don't know why.

apps/desktop/src/app/sync/sources/s3.ts
operramon: I added translations for `Region`.
Not sure if `Access key id` and 'Secret access key' should be translated or not.
In the case of `Bucket` it's the same for all langs.

PR #319, opened 2022-10-18
allow to download external files

ebrehault:


libs/search-widget/src/core/api.ts
```diff
@@ -212,11 +212,16 @@ export const hasAuthData = (): boolean => {
...
 
 export const getFileUrls = (paths: string[]): Observable<string[]> => {
-  if (paths.length === 0 || !isPrivateKnowledgeBox()) {
-    return of(paths.map((path) => `${getRegionalBackend()}${path}`));
-  } else {
-    return getTempToken().pipe(
-      map((token) => paths.map((path) => `${getRegionalBackend()}${path}?eph-token=${token}`)),
-    );
-  }
+  return (paths.length === 0 || !isPrivateKnowledgeBox() ? of('') : getTempToken()).pipe(
```
mpellerin42, line 215: I don't get this: how `return (paths.length === 0 ||…` can return an observable if `paths.length !== 0`?

mpellerin42: ok, I get it now. It took me some time though, which is an indicator that this one liner might deserve to be a two liners?

libs/search-widget/src/old-components/viewer/Metadata.svelte
```diff
@@ -164,29 +159,30 @@
...
-                 closeable
-                 on:cancel={() => showSafariModal = false}
-                 on:confirm={() => showSafariModal = false}>
+  <ConfirmDialog
+    show={showSafariModal}
+    buttons={[{ label: 'Ok', action: 'confirm' }]}
+    closeable
+    on:cancel={() => (showSafariModal = false)}
+    on:confirm={() => (showSafariModal = false)}
+  >
```
mpellerin42, line 183: still those reformatting hitting on us :( 
maybe we should proritize the linter ticket you created some time ago?

libs/search-widget/src/core/api.ts
ebrehault: because I want to write only one pipe

libs/search-widget/src/old-components/viewer/Metadata.svelte
ebrehault: right

PR #325, opened 2022-10-20
[sc-2765] Add labels in bulk from the resource list

mpellerin42:
- Extract label dropdown component and use it in resource list
  - Fix multi-level dropdown
  - Remove checkbox from resource bulk actions section
  - Add label dropdown in resource bulk actions section
- Add selected labels to selected resources


apps/dashboard/src/app/resources/resource-list.component.ts
```diff
@@ -2,12 +2,24 @@ import { ChangeDetectionStrategy, ChangeDetectorRef, Component, OnDestroy, OnIni
...
+  Subject,
+  take,
+} from 'rxjs';
 import { debounceTime, filter, map, switchMap, takeUntil, tap, toArray } from 'rxjs/operators';
 import { ActivatedRoute, Router } from '@angular/router';
 import { LabelValue, Resource, RESOURCE_STATUS, ResourceList, resourceToAlgoliaFormat } from '@nuclia/core';
 import { SDKService, StateService, STFUtils } from '@flaps/core';
-import { SisModalService } from '@nuclia/sistema';
+import { SisModalService, SisToastService } from '@nuclia/sistema';
+import { deDuplicateList } from '../../../../../libs/sdk-core/src/lib/db/resource.helpers';
```
ebrehault, line 22: bad import

ebrehault (approved):

operramon (approved):

apps/dashboard/src/app/resources/resource-list.component.ts
```diff
@@ -325,4 +339,46 @@ export class ResourceListComponent implements OnInit, OnDestroy {
       });
     }
   }
+
+  updateLabelList($event: LabelValue[]) {
+    this.currentLabelList = $event;
+  }
+
+  addLabelsToSelection() {
```
operramon, line 353: I found 2 things, but don't know if they're important:
- The resource modification request is done even if the user doesn't select any label in the dropdown.
- After adding labels to a resource, the "classification" column doesn't update.

mpellerin42: > The resource modification request is done even if the user doesn't select any label in the dropdown.

good point, I'll update the code to prevent that

> After adding labels to a resource, the "classification" column doesn't update. 

The update is triggered but there is a bug in the backend with x-synchronous which makes us doing the refresh too soon. Once fixed on the backend, the front should work as well.

PR #329, opened 2022-10-21
Show message if resource is not processed [sc-2754]

ebrehault:


libs/search-widget/src/old-components/viewer/Viewer.svelte
```diff
@@ -87,6 +89,10 @@
   <div class="viewer-body" class:preview={$showPreview}>
     <div class="viewer-left">
       <InputViewer/>
+      {#if $notProcessed}
+        {$_('error.processing')}
+        <LoadingDots small />
```
mpellerin42, line 94: Do we really want loading dots here? I mean, will the viewer be automatically updated once processed is done? If not, you should not put a loader which imply you can wait for it, it will be there soon

mpellerin42 (approved):

PR #333, opened 2022-10-24
Track dashboard search usage in PostHog [sc-2839]

ebrehault:


apps/dashboard/src/app/components/topbar/topbar.component.ts
```diff
@@ -87,6 +88,7 @@ export class TopbarComponent implements AfterViewInit {
             );
           }
           widget.setActions(actions);
+          widget.addEventListener('search', () => this.trackingService.logEvent('search'))
```
mpellerin42, line 91: I'm not sure to get it: I see here you track the search event only (no query in there), but later on the widget code you dispatched the query with the search. Did you miss to add the query here?

PR #336, opened 2022-10-25
GCS connector

ebrehault:


apps/desktop/src/app/sync/sources/gcs.ts
```diff
@@ -0,0 +1,113 @@
+/// <reference path="../../../../../../node_modules/@types/gapi/index.d.ts" />
+/// <reference path="../../../../../../node_modules/@types/gapi.auth2/index.d.ts" />
```
operramon, line 2: One question, what is it used for ?

operramon (approved):

apps/desktop/src/app/sync/sources/gcs.ts
ebrehault: right, not useful here

operramon (approved):

PR #341, opened 2022-11-02
[sc-2850] Improve account stats line chart

operramon:


apps/dashboard/src/app/components/charts/bar-chart/bar-chart.component.ts
```diff
@@ -98,9 +98,9 @@ export class BarChartComponent implements AfterViewInit, OnDestroy {
     }
 
     // Tooltip event handlers
-    const mouseover = (data: [string, number], index: number, group: any) => {
+    const mouseover = (event: any, data: [string, number]) => {
```
ebrehault, line 101: can't we use `MouseEvent` instead of `any`?

ebrehault (approved):

apps/dashboard/src/app/components/charts/bar-chart/bar-chart.component.ts
operramon: Fixed

PR #345, opened 2022-11-04
Fix loading of pdf.js lib

mpellerin42:


ebrehault (approved):

libs/search-widget/src/tiles/pdf-tile/PdfTile.svelte
```diff
@@ -171,6 +171,19 @@
 <svelte:window bind:innerWidth
                on:keydown={handleKeydown}
                on:resize={(event) => resizeEvent.next(event)}/>
+<svelte:head>
+  <script src="https://cdn.jsdelivr.net/npm/pdfjs-dist@{pdfJsVersion}/build/pdf.min.js"></script>
+  <script src="https://cdn.jsdelivr.net/npm/pdfjs-dist@{pdfJsVersion}/web/pdf_viewer.js"></script>
+  <link  href="https://cdn.jsdelivr.net/npm/pdfjs-dist@{pdfJsVersion}/web/pdf_viewer.css" rel="stylesheet">
```
operramon, line 177: It seems the styles are loaded but are not being applied to the pdf viewer.
It only happens when using the widget as a web component, maybe it's related to the shadow DOM.

mpellerin42: ok, thanks. I'll check again to fix this

PR #378, opened 2022-11-14
[sc-2905] Account dashboard improvements

mpellerin42:
- Add account dashboard item in account navbar
  - Display account type on dashboard
  - Navigate to manage kb/billing by clicking on the corresponding container of the dashboard


ebrehault (approved):

apps/dashboard/src/app/account/account-home/account-home.component.html
```diff
@@ -122,7 +122,20 @@
     </div>
 
     <div class="right-column">
-      <div class="container kb-state">
+      <div
+        class="container plan"
+        routerLink="../billing">
```
operramon, line 127: I remember there's a feature flag to enable/disable the billing page, maybe we should use it here.

PR #384, opened 2022-11-16
Ericbrehault/sc 3011/paragraph split does not work

ebrehault:


libs/sdk-core/src/lib/db/resource/resource.ts
```diff
@@ -107,6 +114,14 @@ export class ReadableResource implements IResource {
       return title;
     }
   }
+
+  getParagraphText(field: IFieldData, paragraph: Paragraph): string | undefined {
```
operramon, line 118: should the sdk changelog be updated ?

ebrehault: right, I forgot 😅

ebrehault: fixed

operramon (approved):

PR #392, opened 2022-11-18
[sc 3035] add suggested label as filter

mpellerin42:
- Remove intent layer in suggestions to have labels directly
- Add suggested label to filters on click
  - Store the filter string directly instead of an encoded filter query
  - Remove the support for encoded filter query from search input
  - Remove `activateFilters` effect and trigger search directly from `searchBy(label: Classification)` function
- Display filters on the right of popup search input

### DIFF
--- a/libs/search-widget/src/common/label/label.utils.ts
+++ b/libs/search-widget/src/common/label/label.utils.ts
@@ -3,12 +3,19 @@ import { isViewerOpen } from '../../core/stores/modal.store';
 import { nucliaStore } from '../../core/old-stores/main.store';
 import { typeAhead } from '../../core/stores/suggestions.store';
 
-export const labelRegexp = new RegExp(/LABEL=\{([^/]+\/[^}]+)}/, 'g');
-export const typingLabelRegexp = new RegExp(/LABEL=?\{?[^}]*}?/, 'g');
-
 export function searchBy(label: Classification) {
   isViewerOpen.set(false);
-  const labelFilter = `LABEL={${label.labelset}/${label.label}}`;
+  const labelFilter = getFilterFromLabel(label);
   typeAhead.set('');
   nucliaStore().filters.next([labelFilter]);
+  nucliaStore().triggerSearch.next();
+}
+
+export function getFilterFromLabel(label: Classification) {
+  return `/l/${label.labelset}/${label.label}`;
+}
+
+export function getLabelFromFilter(filter: string): Classification {
+  const items = filter.split('/');
+  return { labelset: items[2], label: items[3] };
 }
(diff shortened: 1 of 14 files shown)


libs/search-widget/src/common/label/label.utils.ts
```diff
@@ -3,12 +3,19 @@ import { isViewerOpen } from '../../core/stores/modal.store';
...
 export function searchBy(label: Classification) {
   isViewerOpen.set(false);
-  const labelFilter = `LABEL={${label.labelset}/${label.label}}`;
+  const labelFilter = getFilterFromLabel(label);
   typeAhead.set('');
   nucliaStore().filters.next([labelFilter]);
+  nucliaStore().triggerSearch.next();
+}
+
+export function getFilterFromLabel(label: Classification) {
```
ebrehault, line 14: these 2 functions could be in the sdk

libs/search-widget/src/core/old-stores/main.store.ts
```diff
@@ -95,6 +96,23 @@ export const setDisplayedResource = (resource: DisplayedResource) => {
   nucliaStore().displayedResource.next(resource);
 };
 
+export const addLabelFilter = (label: Classification) => {
+  const filter = getFilterFromLabel(label);
+  const currentFilters = nucliaStore().filters.value;
+  nucliaStore().filters.next(currentFilters.concat([filter]));
```
ebrehault, line 102: are we sure it is not possible the label was already in the list?

libs/search-widget/src/core/old-stores/main.store.ts
```diff
@@ -95,6 +96,23 @@ export const setDisplayedResource = (resource: DisplayedResource) => {
...
+};
+
+export const removeLabelFilter = (label: Classification) => {
+  const filter = getFilterFromLabel(label);
+  const currentFilters = nucliaStore().filters.value;
+  const filterIndex = currentFilters.findIndex((f) => f === filter);
```
ebrehault, line 108: a simpler implementation:
```ts
nucliaStore().filters.next(currentFilters.filter((f) => f !== filter));
```

libs/search-widget/src/old-components/search-input/SearchInput.svelte
```diff
@@ -32,26 +32,18 @@
...
   let position: DOMRect | undefined;
   let showSuggestions = false;
+  let hasFilters = false;
   const filters = nucliaState().filters.pipe(
-    tap((filterQuery) => {
-      if (popupSearch) {
-        typeAhead.set(filterQuery.join(''));
+    tap((filters) => {
+      // search box size changes when there are filters or not
+      if (hasFilters !== filters.length > 0) {
```
ebrehault, line 39: that's a bit difficult to understand at first read, what about being more explicit:
```ts
const hasFiltersNow = filters.length > 0;
if (hasFilters !== hasFiltersNow) {
    hasFilters = hasFiltersNow;
…
```

ebrehault (approved):

libs/search-widget/src/core/old-stores/main.store.ts
mpellerin42: good point

operramon (approved):

PR #399, opened 2022-11-23
[sc-3053] Allow to get search results from the outside

operramon:


libs/search-widget/src/core/search-bar.ts
```diff
@@ -24,6 +24,7 @@ export const setupTriggerSearch = (dispatch: (event: string, search: string) =>
         ),
       ),
       switchMap(({ query, options }) => search(query, options)),
+      tap((results) => (dispatch ? dispatch('results', results) : undefined)),
```
mpellerin42, line 27: It's probably a matter of taste, but I don't really like to use conditional ternary operator for this kind of case where we want to call a function or not. On such case I prefer a simple if block (even though it's not a one liner anymore).
```suggestion
      tap((results) => {
        if (typeof dispatch === 'function') {
          dispatch('results', results);
        }
      }),
```

mpellerin42 (approved):

libs/search-widget/src/core/search-bar.ts
operramon: Yes, it was a bit strange.
Fixed.

PR #401, opened 2022-11-25
[sc-3037] Improve performance in large pdf

operramon:


mpellerin42 (approved):

libs/search-widget/src/core/old-stores/viewer.store.ts
```diff
@@ -531,6 +548,12 @@ function findParagraphFromSearchParagraph(
   return paragraphs?.find((paragraph) => text === resource.getParagraphText(field as IFieldData, paragraph));
 }
 
+function hasSpecialChars(resource: Resource) {
```
ebrehault, line 551: So apparently, by using `lengthUnicode` to decide if we can use a basic string slice or not, the performances get better.
As this function is implemented by doing `Array.from(str).length`, it seems that what is costful in `sliceUnicode()` is not to generate the array from the string (the `Array.from(str)` part), but just the fact we make a `slice` + a `join` on this gigantic array.

If that's how it works, I think it would be better to not touch anything in `getParagraphText` or anything and just re-implement `sliceUnicode()` like this:
```ts
export const sliceUnicode = (str: string | undefined, start?: number, end?: number) => {
  if (!str) {
    return '';
  }
  // In JavaScript, '🤖'.length is 2, but all positions in API responses are based on Python
  // and in Python len('🤖') is 1.
  // By converting the string to an array, we can get the correct length and slicing becomes consistent with the API
  // (because the array will split the string into characters, no matter how long they are)
  const arr = Array.from(str);
  const hasSpecialCharacter = arr.length !== str.length
  return hasSpecialCharacter ? arr.slice(start, end).join('') : str.slice(start,end);
};
```

ebrehault (approved):

ebrehault (approved):

PR #417, opened 2022-12-05
[sc-3193] Prevent labels to be half visible on result

operramon:


ebrehault (approved):

libs/search-widget/src/old-components/row/Row.svelte
```diff
@@ -77,7 +77,7 @@
               on:click={() => searchBy(label)} />
           {/each}
           {#if labels.length > 4}
-            <div class="label">+</div>
+            <Label label={{ labelset: labels[4].labelset, label: '+' }} />
```
mpellerin42, line 80: This `+` label looks suspicious to me. Shouldn't this be a button so you can display the hidden labels?

PR #419, opened 2022-12-05
Mix fulltext and semantic results

ebrehault:


libs/search-widget/src/core/old-stores/main.store.ts
```diff
@@ -60,6 +61,71 @@ export const nucliaStore = (): NucliaStore => {
...
+                  smartResults,
+                  resource,
+                  generateFakeParagraphForSentence(allResources, sentence),
+                );
+              }
+            });
+          }
+
+          // add the rest of the fulltext results
+          smartResults = [...smartResults, ...fullTextResults];
```
operramon, line 98: is fullTextResults[0] added again here ?

ebrehault: fixed

libs/search-widget/src/core/old-stores/main.store.ts
```diff
@@ -60,6 +61,74 @@ export const nucliaStore = (): NucliaStore => {
...
+          const semanticResults = results.sentences?.results || [];
+          let smartResults: Search.SmartResult[] = [];
+
+          // best fulltext match goes first
+          if (fullTextResults.length > 0) {
+            smartResults.push(fullTextResults[0]);
+          }
+
+          // if not a keyword search, add the 2 best semantic sentences
+          const looksLikeKeywordSearch = _store!.query.value.split(' ').length < 3;
```
mpellerin42, line 82: is the query in the store trimed already?

libs/search-widget/src/core/old-stores/main.store.ts
```diff
@@ -60,6 +61,74 @@ export const nucliaStore = (): NucliaStore => {
...
+                  generateFakeParagraphForSentence(allResources, sentence),
+                );
+              }
+            });
+          }
+
+          // add the rest of the fulltext results
+          const existingResourceIds = smartResults.map((res) => res.id);
+          const remainingFullTextResults = fullTextResults
+            .slice(1)
```
mpellerin42, line 100: you tested `if (fullTextResults.length > 0)` before adding the first fulltext result to `smartResults`, should you run this test here as well before `fullTextResults.slice(1)`?
I feel like nothing prevent you to arrive here if `fullTextResults.length === 0` (even though I feel like that would be a super corner case)

libs/search-widget/src/widgets/search-video-widget/SearchResults.svelte
```diff
@@ -14,34 +14,12 @@
   import PdfTile from '../../tiles/pdf-tile/PdfTile.svelte';
 
   const showResults = nucliaStore().triggerSearch.pipe(map(() => true));
-  const results = nucliaState().results;
+  const results = nucliaState().smartResults.pipe(tap((results) => console.log(results)));
```
mpellerin42, line 17: console.log?

PR #425, opened 2022-12-08
Cleanup state

ebrehault:


libs/search-widget/src/old-components/embedded-search/EmbeddedSearch.svelte
```diff
@@ -11,7 +11,7 @@
     nucliaStore().triggerSearch.pipe(map(() => true)),
     nucliaState().smartResults.pipe(map((results) => results.length > 0)),
   );
-  const hasQuery = nucliaState().query.pipe(map((query) => !!query));
+  const isEmptyQuery = nucliaState().isEmptyQuery;
```
mpellerin42, line 14: same change should be done on SearchResults widget

PR #430, opened 2022-12-09
[sc-3208] New search page

mpellerin42:
- Fix widget: load label sets in store when filter feature is enabled
- New search page with embedded search widget
- No more search widget on the header
- Little reorganisation of the navbar


apps/dashboard/src/app/components/topbar/topbar.component.ts
```diff
@@ -19,46 +17,16 @@ export class TopbarComponent implements AfterViewInit {
...
-    }),
-    switchMap((kb) => kb.getLabels().pipe(map((labelSets) => ({ kb, labelSets })))),
-    map(({ kb, labelSets }) => {
-      const hasLabels = Object.keys(labelSets).length > 0;
-      const features = !hasLabels
-        ? DEFAULT_FEATURES_LIST.split(',')
-            .filter((feature) => feature !== 'filter')
-            .join(',')
-        : DEFAULT_FEATURES_LIST;
-      return this.sanitized.bypassSecurityTrustHtml(`<nuclia-search id="search-widget" knowledgebox="${kb.id}"
```
ebrehault, line 37: We do want to remove the widget from the topbar, but at the moment that's how we get the viewer working in resources list.
So with this change it will be broken.
You need to inject the widget viewer (which is a custom element too) in resources list page.

ebrehault (changes requested):

ebrehault (approved):

apps/dashboard/src/app/resources/resource-list.component.ts
```diff
@@ -105,6 +108,23 @@ export class ResourceListComponent implements OnInit, OnDestroy {
   labelSets$ = this.sdk.currentKb.pipe(switchMap((kb) => kb.getLabels()));
   currentLabelList: Classification[] = [];
 
+  viewerWidget = this.sdk.currentKb.pipe(
+    distinctUntilKeyChanged('id'),
+    tap(() => {
+      document.getElementById('viewer-widget')?.remove();
+    }),
+    map((kb) =>
+      this.sanitized.bypassSecurityTrustHtml(`<nuclia-viewer id="viewer-widget"
```
operramon, line 117: I have problems when the kb is private, I get 403

mpellerin42: ok, let me have a look

operramon (approved):

PR #448, opened 2022-12-16
[sc-3076] Text tile implementation

mpellerin42:
- `DocumentTile` used in both `PdfTile` and `TextTile`
- Fix navigation in results buttons
- Fix selection state of page indicator


libs/search-widget/src/tiles/tile.utils.ts
```diff
@@ -0,0 +1,5 @@
+export const markRegex = new RegExp(/<\/*mark>/, 'g');
```
ebrehault, line 1: I think NERs can have some attributes in their <mark> tags, like `mark class="ner" family="something">`.
So your regexp should capture more, like:
```/<\/*mark[^>]*>/```

ebrehault (approved):

PR #451, opened 2022-12-19
[sc-3309] Update typography on dashboard

mpellerin42:
Replace `Source sans pro` by Inter (from Google font) in all our apps


libs/search-widget/src/tiles/video-tile/VideoTile.svelte
```diff
@@ -113,6 +113,7 @@
           const summaries = res.summary ? [res.summary] : res.getExtractedSummaries();
           summary = summaries.filter((s) => !!s)[0];
           transcripts = getMainFieldParagraphs(res) as MediaWidgetParagraph[];
+          console.log(transcripts);
```
operramon, line 116: A console log

operramon (approved):

libs/search-widget/src/tiles/video-tile/VideoTile.svelte
mpellerin42: arf yes, good catch!

PR #452, opened 2022-12-19
[sc-3237] Give direct access to Desktop app download

operramon:


apps/dashboard/src/app/knowledge-box/utils.ts
```diff
@@ -7,3 +9,36 @@ export const KB_ROLE_TITLES: { [role in KBRoles]: string } = {
...
+  return null;
+}
+
+export function getDesktopAppUrl(platform: 'mac' | 'win' | 'linux'): Observable<string | null> {
+  const filenames = {
+    mac: /.*-universal-mac\.zip$/,
+    win: /.*\.exe$/,
+    linux: /.*\.snap$/,
+  };
+  return fromFetch('https://api.github.com/repos/nuclia/frontend/releases/latest').pipe(
```
mpellerin42, line 31: you're using this URL here and on `knowledge-box-home.component.ts` as well, you could export it in a const in utils.ts, and use this const everywhere instead.

apps/dashboard/src/app/knowledge-box/utils.ts
```diff
@@ -7,3 +9,36 @@ export const KB_ROLE_TITLES: { [role in KBRoles]: string } = {
...
+  }
+  return null;
+}
+
+export function getDesktopAppUrl(platform: 'mac' | 'win' | 'linux'): Observable<string | null> {
+  const filenames = {
+    mac: /.*-universal-mac\.zip$/,
+    win: /.*\.exe$/,
+    linux: /.*\.snap$/,
+  };
```
mpellerin42, line 30: usually it's better to define regexp once for all (it's better for performances), so you can move this const outside of this function (no need to export it though).

mpellerin42 (approved):

apps/dashboard/src/app/knowledge-box/utils.ts
operramon: fixed

apps/dashboard/src/app/knowledge-box/utils.ts
operramon: fixed 👍

mpellerin42 (approved):

PR #454, opened 2022-12-23
[sc-3146] New sign up flow

mpellerin42:
- New signup page layout
- Signup using sso from Google or GitHub
- Signup with email + password
- Redirect to resource list after sign up flow


operramon (approved):

libs/pastanaga/src/assets/logos/logo.svg
```diff
@@ -1,11 +0,0 @@
-<svg width="93" height="84" viewBox="0 0 93 84" fill="none" xmlns="http://www.w3.org/2000/svg">
```
operramon, line 1: A small thing, changing this logo makes the 404 page look a bit strange.

PR #455, opened 2022-12-23
[sc-3248] Import text resources from a CSV

operramon:


apps/dashboard/src/app/upload/upload-text/upload-text.component.html
```diff
@@ -0,0 +1,74 @@
...
+    <span [innerHtml]="'upload.csv_format' | translate"></span>
+  </p>
+
+  <ncom-hint label="CSV format details">
+    <p></p>
+    <ul>
+      <li>Use commas as field delimiter</li>
+      <li>Fields containing commas or line breaks must be enclosed in double quotes</li>
+      <li>Double quotes appearing inside a quoted field must be prefixed with an additional double quote</li>
+      <li>No header line is required</li>
```
mpellerin42, line 23: This sentence makes me think the CSV doesn't need a header line but could have one. What happens if we include a header line? is it ignored or uploaded in the KB as a resource?
If it is ignore, no change is needed.
If the header line is uploaded as a first resource, you should change the text to something like "Don't include a header line"

mpellerin42 (approved):

apps/dashboard/src/app/upload/upload-text/upload-text.component.html
operramon: If the header is provided it's uploaded as a resource.
The text is fixed now.

PR #468, opened 2023-01-02
[sc-3211] no more status column

mpellerin42:
- Display pending and error resources in their own tables
- Display popover on the pending/failing resources buttons
- Add status column in pending view + improve popover style
- Prevent error when no resources


apps/dashboard/src/app/resources/resource-list.component.html
```diff
@@ -248,16 +278,12 @@
           <th
             cdk-header-cell
             *cdkHeaderCellDef>
-            {{ 'resource.status' | translate }}
+            Status
```
ebrehault, line 281: why no translation here?

ebrehault (approved):

apps/dashboard/src/app/resources/resource-list.component.html
mpellerin42: I just forgot to put it back: this column was removed and then I added it again only on processing tab

ebrehault (approved):

PR #487, opened 2023-01-06
OneDrive connector

ebrehault:


apps/desktop/src/app/upload/select-files/select-files.component.ts
```diff
@@ -60,12 +61,12 @@ export class SelectFilesComponent implements AfterViewInit, OnDestroy {
       concat(
         (this.source as ISourceConnector).getFiles(this.query).pipe(
           catchError((error) => {
-            if (this.source && (error.message = 'Unauthorized' || error.status === 403)) {
+            if (this.source && (this.source.isAuthError || defaultAuthCheck)(error)) {
```
mpellerin42, line 64: are you sure of this syntax? shouldn't it be
```suggestion
            if (this.source && (this.source.isAuthError || defaultAuthCheck(error))) {
```

mpellerin42 (approved):

PR #499, opened 2023-01-11
no sentence labelset + not multivalued

ebrehault:


apps/dashboard/src/app/ontologies/ontology/ontology.component.ts
```diff
@@ -29,13 +29,13 @@ export class OntologyComponent implements OnDestroy {
   ontologyForm = this.formBuilder.group({
     title: ['', [Validators.required, Sluggable()]],
     description: [''],
+    kind: [undefined, [Validators.required]],
```
operramon, line 32: There's an edge case, not sure if it's important:
Some old entity sets have an empty `kind` field. From now on, it won't be possible to modify any field or entity of these sets, because the kind is required but cannot be set.

operramon (approved):

PR #500, opened 2023-01-11
Ericbrehault/sc 3546/brightcove connector

ebrehault:


apps/desktop/src/app/sync/sources/brightcove.ts
```diff
@@ -0,0 +1,142 @@
+import {
```
operramon, line 1: App version should be increased ? or not yet ?

operramon (approved):

apps/desktop/src/app/sync/sources/brightcove.ts
ebrehault: oh right! I always forget…

PR #515, opened 2023-01-17
force validation to be refreshed

ebrehault:


libs/common/src/lib/user/onboarding/onboarding.component.ts
```diff
@@ -79,6 +79,9 @@ export class OnboardingComponent {
       if (enabled) {
         this.onboardingForm.get('phoneInternationalCode')?.setValidators([Validators.required]);
```
operramon, line 80: According to the docs,`addValidators`could be used in case the number format validator should be preserved

operramon (approved):

PR #532, opened 2023-01-25
[sc-3564] Allow to access directly a resource in new widget

operramon:


libs/search-widget/src/widgets/search-video-widget/Tile.svelte
```diff
@@ -0,0 +1,19 @@
+<script lang="ts">
```
ebrehault, line 1: very good idea to refactor this if/else chain in a component

ebrehault (approved):
With a PDF tile, the PDF pages are empty, and I see this in the console:
```
Warning: Indexing all PDF objects, util.js:425
Warning: Invalid stream: "FormatError: Bad FCHECK in flate stream: 120, 194"
```

It works fine on a text tile.

mpellerin42 (approved):

PR #552, opened 2023-02-02
[sc-3370] Reprocess and delete all from failed view

mpellerin42:
- Add a button allowing to select all (not just resources visible on the page) on failure view
- Add a button allowing to reprocess several resources at once
- Add progress indicator for bulk actions


apps/dashboard/src/app/resources/resource-list.component.ts
```diff
@@ -49,6 +53,7 @@ import { SampleDatasetService } from './sample-dataset/sample-dataset.service';
 import { LabelsService } from '../services/labels.service';
 import { PopoverDirective } from '@guillotinaweb/pastanaga-angular';
 import { LOCAL_STORAGE } from '@ng-web-apis/common';
+import { resource } from '../../../../../libs/search-widget/src/core/stores/resource.store';
```
ebrehault, line 56: you probably do not need that here :)

ebrehault (approved):

apps/dashboard/src/app/resources/resource-list.component.ts
mpellerin42: definitely not! good catch :)

operramon (approved):

PR #553, opened 2023-02-02
store job_title after account is created

ebrehault:


libs/common/src/lib/user/onboarding/onboarding.models.ts
```diff
@@ -1,8 +1,8 @@
...
-  industry: string;
-  job_title: string;
+  company?: string;
+  industry?: string;
+  job_title?: string;
   phone?: string;
-  other_search_engines: string;
-  receive_updates: boolean;
+  other_search_engines?: string;
+  receive_updates?: boolean;
```
mpellerin42, line 7: why setting all the fields as optional? As far as I remember all the fields in the form are actually mandatory. Job title is sent later on, but I feel like that should not impact the other fields.

ebrehault: Carles made all the fields optional, that we can call it a first time with all the fields and a second time with job_title only.
We could have imagined a more precise typing but in general I prefer to align with the backend endpoint typing.

mpellerin42: ok

mpellerin42 (approved):

PR #563, opened 2023-02-07
[sc-3846]: label set page improvements

mpellerin42:
- Rename Ontology module/components/route into LabelSet
- Remove feature flag for label-set management
- Fix label-dropdown to keep labels order (don't sort alphabetically)
- Label set page improvements
  - new label set list layout with expander instead of side panel
    - button to edit/delete directly from the label set list
    - list of labels displayed when expander is expanded
  - label set addition/edition form:
    - improved label list UX
    - enable "multiple" checkbox
    - better form validation


apps/dashboard/src/app/label-sets/label-set/label-set.component.ts
```diff
@@ -123,91 +133,61 @@ export class LabelSetComponent implements OnDestroy {
...
-  setMultiple(selected: boolean) {
-    if (this.ontology) {
-      this.ontology.multiple = selected;
-      this.hasChanges = true;
-      this.cdr.markForCheck();
-    }
+    return !!this.labelSet?.labels?.find((label) => label.title === title);
   }
 
   showDuplicationWarning(title: string) {
```
operramon, line 193: A very small thing, text`label-set.eduplicated_label` is misssing

operramon (approved):

apps/dashboard/src/app/label-sets/label-set/label-set.component.ts
mpellerin42: good catch, thanks!!

PR #590, opened 2023-02-16
[sc-3866] Replace viewer with tile widget

mpellerin42:
- Fix dark-stronger color on widget and error management on audio tile
- New viewer store
  - new `getResourceField` method on kb and search-widget api
  - new viewer store for field full id (including resource id), field data and resource title
  - new method `getFieldUrl` extracting field URL from the store
- Update tiles to use new viewer store
- Fixes from UX review
  - update secondary basic button background on hover and active states (one step darker)
  - update expanded tile background color (one step darker)
  - update side panel button color (document tile)
  - update side panel button icon when open (chevrons-right instead of left)
- Factorize common code in Tiles
  - Creation of a common BaseTile managing side panel, search and selection
  - Updating DocumentTile to use BaseTile
  - Creation of MediaTile which is the base from Audio and Video tiles
  - Suppression of stylesheets not needed anymore
- Update viewer widget to display tile
- Add open viewer button from dashboard preview page


libs/sdk-core/src/lib/db/kb/kb.ts
```diff
@@ -115,16 +115,64 @@ export class KnowledgeBox implements IKnowledgeBox {
...
 
+  private _getPath(uuid?: string, slug?: string): string {
+    return !uuid ? `${this.path}/slug/${slug}` : `${this.path}/resource/${uuid}`;
+  }
+
   getResourceFromData(data: IResource): Resource {
     return new Resource(this.nuclia, this.id, data);
   }
 
+  getResourceField(
```
ebrehault, line 131: it should belong to the `ReadableResource` class (and actually there is already a `getField` method, it just does not supports the `show` and `extracted` options but they can be implemented)

mpellerin42: putting this in `ReadableResource` means we need to have a resource object to be able to call this method though, which is not always the case

mpellerin42: There is a `getFields` method in ReadableResource, returning a list of fields, not a `getField` method returning a specific field.

ebrehault (approved):

PR #591, opened 2023-02-17
fix media/non-media check in upload folder

ebrehault:


package.json
```diff
@@ -56,6 +56,7 @@
     "fuse.js": "^3.6.1",
     "googleapis": "^110.0.0",
     "latinize": "^0.5.0",
+    "mime": "^3.0.0",
```
operramon, line 59: one question, the `mime-types` package is already added. Is it necessary to add `mime` package ?

operramon (approved):

package.json
ebrehault: yes because `mime-types` is for NodeJS (and we use it in the desktop), while `mime` works in browser.
Maybe we could replace `mime-types` with `mime`, but it is not a big deal.

mpellerin42 (approved):

PR #592, opened 2023-02-17
[sc-4030] Dataset import proposed when not empty

operramon:
- Use data from `/counter` endpoint to check if the knowledge box is empty (at the moment it's more robust than using`fulltext.facets`)

### DIFF
--- a/apps/dashboard/src/app/resources/resource-list.component.html
+++ b/apps/dashboard/src/app/resources/resource-list.component.html
@@ -368,7 +368,7 @@
 
     <div
       class="pagination"
-      *ngIf="!(isEmpty | async)">
+      *ngIf="resultsLength > 0">
       <stf-pagination
         [page]="page"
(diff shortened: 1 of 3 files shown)


apps/dashboard/src/app/resources/resource-list.component.html
```diff
@@ -368,15 +368,15 @@
 
     <div
       class="pagination"
-      *ngIf="!(isEmpty | async)">
+      *ngIf="resultsLength > 0">
```
ebrehault, line 371: if `isEmpty` is not used anymore, it must be removed from the component implementation

ebrehault (approved):

apps/dashboard/src/app/resources/resource-list.component.html
operramon: Fixed

PR #597, opened 2023-02-23
Ericbrehault/sc 4151/display the generated pdf

ebrehault:


operramon (approved):

libs/search-widget/src/widgets/search-video-widget/SearchBar.svelte
```diff
@@ -104,6 +105,7 @@
     setLang(lang);
 
     setupTriggerSearch(dispatchCustomEvent);
+    loadFieldData();
```
mpellerin42, line 108: loading field data is needed on the results, not really on the SearchBar, why not calling this method in SearchResults?

PR #613, opened 2023-03-03
[sc-4323] Put back full transcripts in media tiles

mpellerin42:


libs/search-widget/src/tiles/base-tile/BaseTile.svelte
```diff
@@ -5,7 +5,7 @@
   import Icon from '../../common/icons/Icon.svelte';
   import type { ResourceField, Search } from '@nuclia/core';
   import type { WidgetParagraph } from '../../core/models';
-  import { PreviewKind } from '../../core/models';
+  import { MediaWidgetParagraph, PreviewKind } from '../../core/models';
```
ebrehault, line 8: `MediaWidgetParagraph` must be imported as `type`

ebrehault (approved):

PR #623, opened 2023-03-17
Ericbrehault/sc 4035/ux for generated answer chat interface

ebrehault:


libs/search-widget/src/core/stores/answers.store.ts
```diff
@@ -1,13 +1,46 @@
...
 }
 
+const EMPTY_ANSWER = { text: '', sources: [] };
 export const answerState = new SvelteState<AnswerState>({
   dialog: [],
+  currentAnswer: EMPTY_ANSWER,
+  isStreaming: false,
 });
+
+export const currentAnswer = answerState.writer<Partial<Answer>>(
```
mpellerin42, line 25: I don't understand why you set the writer as `Partial<Answer>` while you're returning `currentAnswer` which is of type `Answer`.
If you wanted to type the value sent to set the currentAnswer, I think you should have typed your writer as `answerState.writer<Answer, Partial<Answer>>`, no?

libs/search-widget/src/common/expander/Expander.svelte
```diff
@@ -33,7 +34,11 @@
 </script>
 
 <div class="sw-expander">
-  <slot name="header" />
+  <div
+    on:click={() => (expanded = !expanded)}
+    class="header">
+    <Icon name={expanded ? 'chevron-down' : 'chevron-right'} /><slot name="header" />
```
mpellerin42, line 40: To be able to have a nice transition when toggling the expander, it would be better to use only one chevron (let say `chevron-right`) and to add a class on the div when expanded which would transform the icon (like `transform: rotate(90deg)`). Don't forget to add the transition (something like `transition: transform var(--transition-fast)`). 
(I wrote all that from memory, so double check the variable name, but the css variable for transition exists already)

libs/search-widget/src/components/answer/DialogInput.svelte
```diff
@@ -0,0 +1,45 @@
...
+  </div>
+  <input
+    name="nuclia-dialog-field"
+    bind:this={inputElement}
+    {placeholder}
+    tabindex="0"
+    autocomplete="off"
+    autocapitalize="off"
+    spellcheck="false"
+    aria-label="Dialog input"
```
mpellerin42, line 38: should we translate this?

libs/search-widget/src/core/stores/answers.store.ts
```diff
@@ -25,22 +32,30 @@ export const currentAnswer = answerState.writer<Partial<Answer>>(
...
+    currentQuestion: '',
     currentAnswer: EMPTY_ANSWER,
     isStreaming: false,
   }),
 );
 
+export const resetDialog = answerState.writer<void>(
+  () => undefined,
+  (state) => ({ ...state, dialog: state.dialog.slice(0, 1) }),
+);
```
mpellerin42, line 59: why not creating a simple function for this? a writer seems a bit overkill here

ebrehault: I don't get it. How could we change the state if it is not through a `writer`?

libs/search-widget/src/components/answer/InitialAnswer.svelte
```diff
@@ -23,6 +29,12 @@
   </div>
 {/if}
 
+<Modal
+  show={showDialog}
+  closeButton={true}>
+  <Dialog />
```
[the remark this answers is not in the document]
ebrehault: renamed into `Chat`

libs/search-widget/src/components/answer/DialogInput.svelte
```diff
@@ -38,6 +87,14 @@
     aria-label="Dialog input"
     bind:value={question}
     on:keypress={onKeyPress} />
+  {#if $isSpeechEnabled}
+    <div class="microphone">
+      <IconButton
+        icon="microphone"
+        aspect={$isSpeechStarted ? 'solid' : 'basic'}
```
mpellerin42: I guess what you want is to mark the button as active while the speech started. I feel like we should use the active state of the button instead of changing its aspect. 
`IconButton` (and `Button`) component already have an active style which is done by both the `:active` state of the button and an `.active` class. So you can easily add an `active` property on IconButton component to force the active state (it's something we already have on Angular buttons).

libs/search-widget/src/components/answer/ChatInput.svelte
```diff
@@ -0,0 +1,103 @@
...
+    autocapitalize="off"
+    spellcheck="false"
+    aria-label={$_('answer.input.label')}
+    bind:value={question}
+    on:keypress={onKeyPress} />
+  {#if $isSpeechEnabled}
+    <div class="microphone">
+      <IconButton
+        icon="microphone"
+        aspect={$isSpeechStarted ? 'solid' : 'basic'}
```
mpellerin42, line 95: I guess what you want is to mark the button as active while the speech started. I feel like we should use the active state of the button instead of changing its aspect.
IconButton (and Button) component already have an active style which is done by both the :active state of the button and an .active class. So you can easily add an active property on IconButton component to force the active state (it's something we already have on Angular buttons)

libs/search-widget/src/core/stores/answers.store.ts
mpellerin42: you can use the `update` function of your state. See in `annotations` store: we have our store, we have a few writers for the things we read and write, and we have a few functions allowing to just update specific things of the store.

mpellerin42 (approved):
looks good to me

PR #627, opened 2023-03-20
Apply file size restriction in desktop

ebrehault:


apps/desktop/src/app/sync/destinations/nuclia-cloud.ts
```diff
@@ -56,10 +57,18 @@ class NucliaCloudKBImpl implements IDestinationConnector {
...
-      return kb$.pipe(
+      const mimetype = lookup(filename) || 'application/octet-stream';
+      return this.getAccount().pipe(
+        tap((account) => {
+          const applicableLimit = this.isMedia(mimetype)
+            ? account.limits.upload.upload_limit_max_media_file_size
+            : account.limits.upload.upload_limit_max_non_media_file_size;
+          if (blob.size > applicableLimit) {
+            console.error(`File too large. Size=${blob.size}, limit=${applicableLimit}`);
+            throw new Error(`File "${filename}" is too large.`);
```
mpellerin42, line 68: shouldn't we display a toast or something for showing the details of what failed (here the size limit) not only on the console (specially because I guess the console is not really accessible in electron)?

operramon (approved):

PR #633, opened 2023-03-22
Ericbrehault/sc 4607/move from search to find

ebrehault:


libs/sdk-core/src/lib/db/search/search.ts
```diff
@@ -2,6 +2,45 @@ import { catchError, map, of, tap } from 'rxjs';
...
+  params['features'] = features;
+  const { inTitleOnly, ...others } = options || {};
+  if (inTitleOnly) {
+    params['fields'] = ['a/title'];
+  }
+
+  params['shards'] = nuclia.currentShards?.[kbid] || [];
+
+  const searchMethod = useGet
+    ? nuclia.rest.get<Search.FindResults | { detail: string }>(`${path}/find?${serialize(params, others)}`)
```
mpellerin42, line 29: I feel like it would worth having a model `ErrorResponse` for this `{ detail: string }`, because it's actually common for all the endpoints, don't you think?

libs/sdk-core/src/lib/db/search/chat.ts
```diff
@@ -0,0 +1,39 @@
...
+        const sourcesData = data.slice(4, sourcesLength + 4);
+        sources = JSON.parse(atob(new TextDecoder().decode(sourcesData.buffer)));
+      }
+      if (sources && data.length > sourcesLength + 4) {
+        text = new TextDecoder().decode(data.slice(sourcesLength + 4).buffer);
+        end = text.includes(END_OF_STREAM);
+        if (end) {
+          text = text.split(END_OF_STREAM)[0];
+        }
+      }
```
mpellerin42: it would be clearer if you added some comments to this algo. This check of `data.length > 4` or `data.length > sourcesLength + 4` is quite cryptic

libs/sdk-core/src/lib/rest/rest.ts
```diff
@@ -148,4 +148,41 @@ export class Rest implements IRest {
...
+    return new Observable<Uint8Array>((observer) => {
+      fetch(path, { method: 'POST', headers: this.getHeaders('POST', path), body: JSON.stringify(body) }).then(
+        (res) => {
+          const reader = res.body?.getReader();
+          if (!reader) {
+            observer.next(new Uint8Array());
+            observer.complete();
+          } else {
+            let data = new Uint8Array();
+            const readMore = () => {
```
mpellerin42, line 163: why did you create this function? why not directly running the code it contains?

libs/search-widget/src/core/utils.ts
```diff
@@ -208,6 +208,27 @@ export function mapSmartParagraph2WidgetParagraph(paragraph: Search.FindParagrap
   } as WidgetParagraph;
 }
 
+export function mapParagraph2SmartParagraph(paragraph: Search.Paragraph): Search.FindParagraph {
```
mpellerin42, line 211: is this function coming from somewhere else? if so we should delete the code from the other place to avoid duplicated code.

libs/common/src/assets/i18n/en.json
```diff
@@ -687,6 +687,8 @@
 	"stash.valid_emails_label": "Emails",
 	"stash.view_edit_service_access": "View / Edit existing Service Access",
 	"stash.widgets": "Widgets",
+	"stash.widgets.answers": "Answer generation",
+	"stash.widgets.answers.help": "Generate answer to user questions and allow to chat with the system",
```
mpellerin42, line 691: `Generate answer to user questions` seems weird, no ? Did you mean `Generate answer from user questions` ?

libs/search-widget/src/components/answer/Answer.svelte
```diff
@@ -0,0 +1,37 @@
...
+    </div>
+    <Expander>
+      <h3
+        class="title-xs"
+        slot="header">
+        {$_('answer.sources')}
+      </h3>
+      <div class="results">
+        {#each sources as result}
+          <Tile {result} />
```
operramon, line 28: The sources tiles are displayed always as text tiles. Is this expected ?

libs/sdk-core/src/lib/db/search/search.ts
ebrehault: Right, we will do that after this refactoring.

libs/sdk-core/src/lib/rest/rest.ts
ebrehault: because this function calls itself recursively

libs/search-widget/src/core/utils.ts
ebrehault: no, I did it myself, I did not copy, I swear

libs/search-widget/src/components/answer/Answer.svelte
ebrehault: Not always but yes in many case (when the best paragraph is coming from the title).
Fixed.

libs/sdk-core/src/lib/rest/rest.ts
mpellerin42: right, I missed one call

libs/search-widget/src/widgets/search-widget/SearchResults.svelte
```diff
@@ -55,6 +55,7 @@
     }
     loadFonts();
     loadSvgSprite().subscribe((sprite) => (svgSprite = sprite));
+    entityRelations.subscribe((en) => console.log(en));
```
mpellerin42, line 58: did you forget something here?

libs/search-widget/src/components/answer/Chat.scss
```diff
@@ -51,6 +51,9 @@
       grid-template-columns: var(--rhythm-10) auto var(--rhythm-10);
       margin: auto 0 var(--rhythm-4) 0;
     }
+    .answer {
+      width: calc(var(--resource-modal-width-md, fit-content) - var(--rhythm-10) - var(--rhythm-16));
```
mpellerin42, line 55: this is super weird: `fit-content` is there for to fit the content of the div, so substracting a value to it would probably make the width too small… Also, the variable `--resource-modal-width-md` is meant to be used on modals, but as far as I know the chat is like a tile, so it is fullscreen when expanded, no?

libs/search-widget/src/core/stores/effects.ts
```diff
@@ -123,26 +123,23 @@ export function initAnswer() {
...
         filter((answer) => !!answer),
         map((answer) => (answer as Chat.Answer).text),
         distinctUntilChanged(),
-        switchMap((answer) =>
-          isSpeechOn.pipe(
-            filter((on) => on),
-            map(() => answer),
-          ),
-        ),
+        tap((answer) => console.log(answer)),
```
mpellerin42, line 142: console.log

PR #673, opened 2023-04-07
Fix Chat layout [sc-4853]

mpellerin42:


libs/search-widget/src/components/answer/Answer.svelte
```diff
@@ -5,28 +5,38 @@
...
   export let rank = 0;
+  export let hideFeedback = false;
   let text = '';
+  let innerWidth = window.innerWidth;
 
+  const dispatch = createEventDispatcher();
   const NEWLINE = new RegExp(/\n/g);
-
   $: text = answer.text?.replace(NEWLINE, '<br>') || '';
+  $: isMobile = innerWidth < 448;
```
ebrehault, line 18: next time we need this somehwere, it should be in a `utils`

ebrehault (approved):

libs/search-widget/src/components/answer/Answer.svelte
ebrehault: the method

PR #684, opened 2023-04-12
[sc-4459] Add an icon to copy the code on hint component

mpellerin42:
Hint component refactoring so:
- it displays a copy button when there is some code (ie `pre>code` elements) present
- it has a new input `learnMore` to show a generic learn more link.

### DIFF
--- a/libs/common/src/lib/hint/hint.component.ts
+++ b/libs/common/src/lib/hint/hint.component.ts
@@ -37,8 +37,6 @@ export class HintComponent implements AfterContentInit, AfterViewInit, OnChanges
 
   @ViewChild('content') content?: ElementRef;
   @ViewChild('container') container?: ElementRef;
-  @ViewChild('codeExample') codeExample?: ElementRef;
-  @ViewChild('instructions') instructions?: ElementRef;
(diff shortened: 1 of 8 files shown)


libs/common/src/lib/hint/hint.component.ts
```diff
@@ -42,21 +35,26 @@ export class HintComponent implements AfterContentInit, AfterViewInit, OnChanges
   }
   private _noMaxWidth = false;
 
-  isExpanded = false;
-
   @ViewChild('content') content?: ElementRef;
   @ViewChild('container') container?: ElementRef;
+  @ViewChild('codeExample') codeExample?: ElementRef;
+  @ViewChild('instructions') instructions?: ElementRef;
```
operramon, line 41: These two viewChild are used ?

operramon (approved):

libs/common/src/lib/hint/hint.component.ts
mpellerin42: not anymore, I forgot to remove them. I'll clean up, thanks

PR #704, opened 2023-04-14
[sc-5055] Payment Improvements

operramon:
- A tier can be selected in the budget calculator so that the estimation is more accurate
- Add the enterprise tier information
- Add a modal to review the data before subscribing
- Add email field in the customer form
- Support 3d secure authentication. 
- Change customer form to allow customers without company


apps/dashboard/src/app/account/billing/calculator/calculator.component.html
mpellerin42, line 1: I see you used a `pa-modal-dialog` here, while it contains a form which seems to be quite long. `pa-modal-advanced` might be a better choice when you have a long form to display, because it is bigger than the dialog one. It also comes with a close button by default. That being said, I didn't checkout the design, so don't feel you have to change it, I just wanted you two know about the two types of modals Pastanaga provides.

apps/dashboard/src/app/account/billing/checkout/checkout.component.html
```diff
@@ -1,7 +1,33 @@
...
     <h2>{{ 'billing.payment_details' | translate }}</h2>
-    <div class="body-m">{{ 'billing.payment_details.description' | translate }}</div>
+  </div>
+
+  <div>
+    <div class="section-header">
+      <div class="section-title">
+        <pa-icon
+          name="circle-check"
+          color="#f06"></pa-icon>
```
mpellerin42, line 11: FYI you can also change icon color simply by setting the color attribute in CSS like 
```scss
.section-title pa-icon {
  color: $color-primary-regular;
}
```

mpellerin42 (approved):

apps/dashboard/src/app/account/billing/calculator/calculator.component.html
operramon: Thanks for the info, I'll check

apps/dashboard/src/app/account/billing/checkout/checkout.component.html
mpellerin42: Basically, `pa-icon` inherit its color from its parent. You could also do
```scss
.section-title {
  color: $color-primary-regular;
}
```
if everything in `.section-title` shared the same color.

ebrehault (approved):

PR #718, opened 2023-04-26
Ericbrehault/sc 4424/upload service in a decoupled process rebased

ebrehault:


apps/desktop/src/app/main-layout/sidebar.component.html
```diff
@@ -1,59 +1,69 @@
...
-          <span (click)="goToHistory(true)">{{ 'history.active' | translate }}</span>
+        <li>
+          <span (click)="goToHistory()">{{ 'history.title' | translate }}</span>
         </li>
+        <ul>
+          <li>
+            <span (click)="goToHistory(true)">{{ 'history.active' | translate }}</span>
+          </li>
+        </ul>
       </ul>
```
mpellerin42, line 66: this list within a list with only one `li` each seems super weird to me. What do you try to achieve here?

apps/desktop/src/app/sync/dynamic-connector.ts
```diff
@@ -59,6 +56,11 @@ export class DynamicConnectorWrapper implements ISourceConnector {
     }
   }
 
+  getParametersValues(): ConnectorParameters {
+    //TODO
```
mpellerin42, line 60: is this something you forgot or which is planned for later?

apps/desktop/src/app/sync/sources/brightcove.ts
```diff
@@ -9,15 +9,15 @@ import {
...
-  description: 'Video delivery platform',
-  factory: () => of(new BrightcoveImpl()),
-};
+// export const BrightcoveConnector: SourceConnectorDefinition = {
+//   id: 'brightcove',
+//   title: 'Brightcove',
+//   logo: 'assets/logos/brightcove.svg',
+//   description: 'Video delivery platform',
+//   factory: () => of(new BrightcoveImpl()),
+// };
```
mpellerin42, line 18: dead code

apps/desktop/src/app/sync/sources/dropbox.ts
```diff
@@ -39,104 +30,17 @@ class DropboxImpl implements ISourceConnector {
...
-  /* eslint-disable  @typescript-eslint/no-explicit-any */
-  private filterFiles(raw: any): boolean {
-    return raw?.['.tag'] !== 'folder';
-  }
-
-  /* eslint-disable  @typescript-eslint/no-explicit-any */
-  private filterResults(raw: any): boolean {
-    return raw.match_type?.['.tag'] !== 'folder';
+  goToOAuth(reset?: boolean) {
+    return of(true);
```
mpellerin42, line 40: I don't really understand what's the point of this method

apps/desktop/src/app/sync/sources/dropbox.ts
```diff
@@ -39,104 +30,17 @@ class DropboxImpl implements ISourceConnector {
...
-      fetch(`https://content.dropboxapi.com/2/files/download`, {
-        method: 'POST',
-        headers: {
-          Authorization: `Bearer ${localStorage.getItem(TOKEN || '')}`,
-          'Dropbox-API-Arg': JSON.stringify({ path: resource.originalId }),
-        },
-      }).then((res) => res.blob()),
-    );
+  authenticate(): Observable<boolean> {
+    return of(!!this.getParametersValues().token);
```
mpellerin42, line 44: seems like this method is checking if the connector is authenticated, so the name `authenticate` is misleading

apps/desktop/src/app/sync/sources/folder.ts
```diff
@@ -12,15 +12,15 @@ import { Observable, BehaviorSubject, of, throwError } from 'rxjs';
...
-  description: 'Upload a folder from your device',
-  factory: () => of(new FolderImpl()),
-};
+// export const FolderConnector: SourceConnectorDefinition = {
+//   id: 'folder',
+//   title: 'Folder',
+//   logo: 'assets/logos/folder.svg',
+//   description: 'Upload a folder from your device',
+//   factory: () => of(new FolderImpl()),
+// };
```
mpellerin42, line 21: dead code

apps/desktop/src/app/sync/sources/gcs.ts
```diff
@@ -14,16 +14,16 @@ const BUCKET_KEY = 'GCS_BUCKET';
...
-  factory: () => of(new GCSImpl()),
-};
+// export const GCSConnector: SourceConnectorDefinition = {
+//   id: 'gcs',
+//   title: 'Google Cloud',
+//   logo: 'assets/logos/gcs.svg',
+//   description: 'File storage service developed by Google',
+//   helpUrl: 'https://docs.nuclia.dev/docs/batch/nda/#google-drive-and-google-cloud-connectors-usage',
+//   factory: () => of(new GCSImpl()),
+// };
```
mpellerin42, line 24: dead code?

apps/desktop/src/app/sync/sources/gdrive.ts
```diff
@@ -1,20 +1,20 @@
...
-  factory: () => of(new GDriveImpl()),
-};
+// export const GDrive: SourceConnectorDefinition = {
+//   id: 'gdrive',
+//   title: 'Google Drive',
+//   logo: 'assets/logos/gdrive.svg',
+//   description: 'File storage and synchronization service developed by Google',
+//   helpUrl: 'https://docs.nuclia.dev/docs/batch/nda/#google-drive-and-google-cloud-connectors-usage',
+//   factory: () => of(new GDriveImpl()),
+// };
```
mpellerin42, line 15: dead code?

apps/desktop/src/app/sync/sources/s3.ts
```diff
@@ -13,15 +13,15 @@ import { Observable, of, from, map } from 'rxjs';
...
-  description: 'Object storage service developed by Amazon',
-  factory: () => of(new S3Impl()),
-};
+// export const S3Connector: SourceConnectorDefinition = {
+//   id: 's3',
+//   title: 'AWS S3',
+//   logo: 'assets/logos/s3.svg',
+//   description: 'Object storage service developed by Amazon',
+//   factory: () => of(new S3Impl()),
+// };
```
mpellerin42, line 22: dead code?

apps/desktop/src/app/sync/sync.service.ts
```diff
@@ -72,56 +65,62 @@ export class SyncService {
...
-    const queue: Sync[] = JSON.parse(localStorage.getItem(QUEUE_KEY) || '[]');
-    Object.values(queue).forEach((sync) => {
-      if (!sync.completed) {
-        sync.started = false;
-      }
-    });
-    this._queue = queue;
-    this.onQueueUpdate();
-    this.fetchDynamicConnectors();
+    // this.fetchDynamicConnectors();
```
mpellerin42, line 103: dead code?

apps/desktop/src/app/sync/sync.service.ts
```diff
@@ -72,56 +65,62 @@ export class SyncService {
...
+    of(true)
+      .pipe(
+        filter(() => !!this._syncServer.getValue()),
```
mpellerin42, line 106: why starting by an `of(true` instead of starting directly by `this._syncServer`?
```suggestion
this._syncServer
      .pipe(
      take(1),
        filter((value) => !!value),
```

apps/desktop/src/app/sync/sync.service.ts
```diff
@@ -392,4 +296,95 @@ export class SyncService {
...
+    return localStorage.getItem(SOURCE_NAME_KEY) || '';
+  }
+
+  setCurrentSourceId(id: string) {
+    if (id) {
+      localStorage.setItem(SOURCE_NAME_KEY, id);
+    } else {
+      localStorage.removeItem(SOURCE_NAME_KEY);
+    }
+  }
```
mpellerin42, line 379: do you expect an empty string? the signature tends to feel like you will never get inside the else block.

libs/sdk-core/src/lib/rest/rest.ts
```diff
@@ -86,24 +85,16 @@ export class Rest implements IRest {
...
-        if (!response.ok) {
-          return throwError(() => response);
+    return from(
+      fetch(this.getFullUrl(path), {
+        method,
+        headers: this.getHeaders(method, path, extraHeaders, synchronous),
+        body: specialContentType ? body : JSON.stringify(body),
+      }).then((res) => {
+        if (!res.ok) {
+          throw new Error(`${res.status}`);
```
mpellerin42, line 95: why do you return only the status in the Error? I a lot of cases we don't use error details, but there are some cases (like in the manager) where having the details would be valuable. I feel like here you're removing too much informations from the error and I don't really understand what's the point, why not throwing the actual error (and so `res` entirely I guess).

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "nuclia",
-  "version": "1.2.4",
+  "version": "2.0.0",
```
mpellerin42, line 3: Good to update the version, but you forgot to update the CHANGELOG

mpellerin42 (changes requested):

apps/desktop/electron/src/service/connectors/dropbox.ts
operramon, line 1: I cannot search files in the Dropbox connector but I don't know why.
Is there a way to debug the dev express server ?

apps/desktop/electron/src/service/dynamic-connectors.ts
```diff
@@ -0,0 +1,40 @@
...
+  if (!fs.existsSync(connectorsPath)) {
+    return;
+  }
+  fs.readdirSync(connectorsPath).forEach((file) => {
+    const filePath = path.join(connectorsPath, file);
+    import(filePath).then((module) => {
+      // console.log(module);
+    });
+  });
+};
```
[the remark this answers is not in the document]
ebrehault: I was considering using a feature flag to enable/disable dynamic connectors but it was a bit of effort, so for now it is just commented in sync.ts.
I will add a comment to mention it is still useful but disabled at the moment.

apps/desktop/electron/src/service/sources.ts
```diff
@@ -0,0 +1,270 @@
...
+  sinceGMT?: string,
+  folders?: SyncItem[],
+): Observable<{ success: boolean; results: SyncItem[]; error?: string }> {
+  const connector = getSourceInstance(sourceId);
+  try {
+    return connector.getLastModified(sinceGMT || '2000-01-01T00:00:00.000Z', folders).pipe(
+      map((results) => ({ success: true, results })),
+      catchError((err) => {
+        console.error(`Error on ${sourceId}: ${err.message}`);
+        return of({ success: false, results: [], error: `${err}` });
```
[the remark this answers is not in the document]
ebrehault: Depending on the connector, the error object can be very different, and we have poor control on it (mostly with dynamic connectors) that way I am sure I always get a string.

apps/desktop/electron/src/service/sources.ts
```diff
@@ -0,0 +1,270 @@
...
+      }),
+    );
+  } catch (err) {
+    return of({ success: false, results: [], error: `${err}` });
```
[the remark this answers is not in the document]
ebrehault: same

apps/desktop/electron/src/service/sources.ts
```diff
@@ -0,0 +1,270 @@
...
+export function getLogs(): LogRow[] {
+  return LOG;
+}
+
+export function addLog(id: string, source: Source, count: number, errors: string) {
+  LOG.push({
+    from: id,
+    to: source.kb?.knowledgeBox || 'Unknown kb',
+    errors,
+    progress: 100,
```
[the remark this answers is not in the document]
ebrehault: Yes, this method store past events, so they are always finished.
The `progress` is only useful with active sync logs.

apps/desktop/electron/src/service/sync.ts
```diff
@@ -0,0 +1,145 @@
...
+  incrementActiveSyncLog,
+  addLog,
+  removeActiveSyncLog,
+} from './sources';
+// import { importConnector, loadConnectors } from './dynamic-connectors';
+import { concatMap, delay, map, of, switchMap, tap, toArray } from 'rxjs';
+
+export const sync = () => {
+  // importConnector('https://nuclia.github.io/status/connectors/youtube.js');
+  // loadConnectors();
```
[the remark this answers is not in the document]
ebrehault: I will had a comment to explain that's how dynamic connectors are disabled for now

apps/desktop/src/app/main-layout/sidebar.component.html
ebrehault: That was the existing structure, I have only changed the `(click)` here, but my understanding is it is a trick to implement the indendation.

apps/desktop/src/app/sync/dynamic-connector.ts
ebrehault: dynamic connectors are off for now, I need to put some thinking into it, they will be implemented in a future PR

apps/desktop/src/app/sync/sources/brightcove.ts
ebrehault: All the connectors but OneDrive and DropBox haven't been migrated to the new system, I want to keep the code because the way they use their expected API is relevant.

apps/desktop/src/app/sync/sources/dropbox.ts
ebrehault: All the connectors must implement a certain behaviour, but depending on how they work, some of the expected methods do nothing (typically, we do not use OAuth on Dropbox)

apps/desktop/src/app/sync/sources/dropbox.ts
ebrehault: Here again some connectors need to run some code to authenticate, this is where it is supposed to happen, but for DropBox there is nothing to do, so as far as the token is defined, we can continue.

apps/desktop/src/app/sync/sources/folder.ts
ebrehault: Only Dropbox and OneDrive are relevant for now, I just want to keep the old code

apps/desktop/src/app/sync/sources/gcs.ts
ebrehault: same

apps/desktop/src/app/sync/sources/gdrive.ts
ebrehault: same

apps/desktop/src/app/sync/sources/s3.ts
ebrehault: same

apps/desktop/src/app/sync/sync.service.ts
ebrehault: I will had a comment

apps/desktop/src/app/main-layout/sidebar.component.html
mpellerin42: this structure is a non sense. This is semantically wrong and can be easily achieved using pure css.

libs/sdk-core/src/lib/rest/rest.ts
ebrehault: As the backend uses a status for a unique reason on each endpoint, we do not loose any information by just keeping the status.

Moreover, error management was really hell in the desktop app and, psychologically, I am not ready to invest time to improve it now.

So let's address that in a future task if at some point we think having more detail errors is useful.

package.json
ebrehault: there is no CHANGELOG in the root (that's not the SDK's package here, but the main one)

apps/desktop/electron/src/service/connectors/dropbox.ts
ebrehault: yes you can run it like that: ```yarn desktop-server-dev``` and run the UI in the browser with ```nx serve desktop```
(I will mention that in README actually).

Search is working fine for me, did you use a fresh token?

mpellerin42 (approved):

apps/desktop/electron/src/service/connectors/dropbox.ts
operramon: yes the token is new. I'll investigate it.

PR #719, opened 2023-04-27
Fix label on budget input [sc-5217]

mpellerin42:


apps/dashboard/src/app/account/billing/checkout/checkout.component.html
```diff
@@ -294,6 +294,7 @@ <h2>{{ 'billing.payment_details' | translate }}</h2>
         <div class="budget-input">
           <pa-input
             type="number"
+            externalLabel
```
operramon, line 297: I'll keep it in mind

operramon (approved):

PR #728, opened 2023-05-02
Ericbrehault/sc 5283/fix dataset import

ebrehault:


libs/common/src/lib/resources/sample-dataset/dataset-selector/dataset-selector.component.ts
```diff
@@ -54,7 +54,8 @@ export class DatasetSelectorComponent {
         loadingModal.close();
         this.imported.emit(true);
       },
-      error: () => {
+      error: (err) => {
+        console.log(err);
```
mpellerin42, line 58: do you want to keep this console.log?

mpellerin42 (approved):

libs/common/src/lib/resources/sample-dataset/dataset-selector/dataset-selector.component.ts
ebrehault: whih console.log? 😇

operramon (approved):

PR #735, opened 2023-05-02
[sc-5116] view the knowledge graph

mpellerin42:


libs/search-widget/src/components/knowledge-graph/Graph.svelte
```diff
@@ -0,0 +1,186 @@
+<script lang="ts">
```
ebrehault, line 1: We do not want a direct dependency to d3 because it will increase the size of the widget bundle with unneeded stuff in many use cases.
We should load it from a public CDN (as we do for pdf.js for example).

As I need d3 in the Britinsurance demo, here is a possible implementation but feel free to adapt/re-implement:
https://github.com/nuclia/frontend/blob/ericbrehault/sc-5269/britinsurance-pov/libs/search-widget/src/components/charts/PieChart.svelte#L152

libs/search-widget/src/components/knowledge-graph/KnowledgeGraph.svelte
```diff
@@ -0,0 +1,257 @@
+<script lang="ts">
+  import { fieldData } from '../../core/stores/viewer.store';
+  import Expander from '../../common/expander/Expander.svelte';
+  import { translateInstant } from '../../core/i18n';
+  import { generatedEntitiesColor } from '../../core/utils';
+  import Graph from './Graph.svelte';
+  import { extent, forceLink, forceManyBody, forceX, forceY, SimulationLinkDatum } from 'd3';
```
ebrehault, line 7: use a cdn

libs/search-widget/src/components/knowledge-graph/KnowledgeGraphPanel.svelte
```diff
@@ -0,0 +1,68 @@
...
+          </Button>
+        </div>
+      {/if}
+    </section>
+
+    {#if $graphSelectionParagraphs.length > 0}
+      <section class="results-section">
+        <ul>
+          {#each $graphSelectionParagraphs as paragraph, index}
+            <!--            <li>{paragraph.text}</li>-->
```
operramon, line 54: is it old code ?

libs/search-widget/src/components/knowledge-graph/Graph.svelte
operramon, line 1: don't know if this is expected: the first time the graph is rendered it doesn't have nodes, only links. Then after clicking a family checkbox the nodes are rendered.

operramon (approved):

libs/search-widget/src/components/knowledge-graph/Graph.svelte
mpellerin42: It is not expected. I battle the opposite bug (where I had first only nodes) and didn't notice this bug with only links because of auto-reload. I'll work on that.

libs/search-widget/src/components/knowledge-graph/KnowledgeGraphPanel.svelte
mpellerin42: probably yes, thanks, I'll cleanup

libs/search-widget/src/components/knowledge-graph/Graph.svelte
mpellerin42: Right, I'll do the changes

ebrehault (approved):

PR #792, opened 2023-05-25
Api key page should be accessible to kb owner

mpellerin42:


apps/dashboard/src/app/app-routing.module.ts
```diff
@@ -160,7 +160,7 @@ const routes: Routes = [
               {
                 path: 'keys',
                 component: KnowledgeBoxKeysComponent,
-                canActivate: [accountOwnerGuard],
+                canActivate: [knowledgeBoxOwnerGuard],
```
operramon, line 163: I just tried to create a API key with a user that is kb owner (but not account owner). It seems it's not allowed by the backend, it returns 403.

mpellerin42: Thanks @operramon , I will create a ticket for them as well then.

PR #797, opened 2023-05-30
[sc-5732] sitemap connector

mpellerin42:


apps/desktop/electron/src/service/sitemap-parser.ts
```diff
@@ -0,0 +1,52 @@
...
+      resolve([...urls, ...sitemaps.flat()]);
+    });
+  });
+}
+
+export function getSiteMap(url: string): Observable<SiteMapModel[]> {
+  return from(fetchSitemap(url)).pipe(switchMap((content) => from(parseSitemap(content))));
+}
+
+// const sitemapUrl = 'https://nuclia.com/sitemap.xml';
```
ebrehault, line 41: to be cleaned, no?

package.json
ebrehault, line 1: you need to update the version to get a new desktop app release

ebrehault (approved):

apps/desktop/electron/src/service/sitemap-parser.ts
mpellerin42: indeed, thanks

operramon (approved):

PR #804, opened 2023-06-02
[sc-5821] Reset search pagination

operramon:


ebrehault (approved):

libs/search-widget/src/core/search-bar.ts
```diff
@@ -37,6 +38,7 @@ export const setupTriggerSearch = (
             switchMap(() => searchQuery.pipe(take(1))),
             tap((query) => (dispatch ? dispatch('search', query) : undefined)),
             tap((query) => (!trigger?.more ? ask.next({ question: query, reset: true }) : undefined)),
+            tap(() => (!trigger?.more ? pageNumber.set(0) : undefined)),
```
mpellerin42, line 41: In your code, it seems pageNumber can be only `0` or `undefined`, but what when we scroll down to get next page?

operramon: the code maybe is confusing, when scrolling down it doesn't set the page number as trigger.more is `true`

mpellerin42: yes, I got that, but this is actually my point: pageNumber is updated only here on the code, so currently pageNumber's value can be either `0`, either `undefined`. I'm wondering if it should be updated when we load the next pages?

operramon: oh, seems the page number increment is here:
https://github.com/nuclia/frontend/blob/oriolperramon8396/sc-5821/reset-search-pagination/libs/search-widget/src/core/stores/search.store.ts#L137

mpellerin42: oh! this is the part I missed! perfect then :)

mpellerin42 (approved):

PR #809, opened 2023-06-06
[sc-5763] adjust manager app to address latest features

mpellerin42:
- Field indexer_slow_replicas has been deprecated and should be replaced for the max_dedicated_processors field
- it would be necessary to review that all those fields corresponding to numerical values ​​must be sent as integers instead of being sent as strings
- Limits fields `monthly_limit_chars_processed` and `monthly_limit_non_media_files_processed` are no longer needed, so both could be removed from the limits form
- rethink and implement a different way to set the limits in the limits form (can be `number > 0`,` -1` or `null`)
- provide an option to reset all limits to default values OR unlimited? (this will probably what Eudald uses the most)
- add and manage optional `trial_expiration_date` on account details that can be updated.
- The amount of blocked features has been extended due to the need to perform more granular blocking operations, UI must be updated accordingly
- A TRIAL account should be upgradable to ENTERPRISE, and enterprise can be downgraded to trial. No other changes allowed
- `max_kbs` as same behaviour as limits: 3 possible values (`-1` for unlimited, `null` for default value, and `number > 0` for specified limit).

### DIFF
--- a/apps/manager-v2/src/app/manage-accounts/account-details/limits/limits.component.ts
+++ b/apps/manager-v2/src/app/manage-accounts/account-details/limits/limits.component.ts
@@ -16,7 +16,6 @@ export class LimitsComponent implements OnInit, OnDestroy {
   private unsubscribeAll = new Subject<void>();
 
   limitsForm = new FormGroup({
-    processing: new FormGroup({}),
     upload: new FormGroup({}),
     usage: new FormGroup({}),
   });
@@ -34,12 +33,6 @@ export class LimitsComponent implements OnInit, OnDestroy {
       .getAccount()
       .pipe(takeUntil(this.unsubscribeAll))
       .subscribe((accountDetails) => {
-        Object.entries(accountDetails.limits.processing || {}).forEach(([key, limit]) => {
-          this.limitsForm.controls.processing.addControl(
-            key,
-            new FormControl<number>(limit, { nonNullable: true, validators: [Validators.required] }),
-          );
-        });
         Object.entries(accountDetails.limits.upload || {}).forEach(([key, limit]) => {
           this.limitsForm.controls.upload.addControl(
             key,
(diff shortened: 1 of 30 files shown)


apps/manager-v2/src/app/manage-accounts/account-details/limits/limits.component.ts
```diff
@@ -32,69 +35,152 @@ export class LimitsComponent implements OnInit, OnDestroy {
...
+          return value;
+        }, {} as { [key: string]: number }) as any;
+        return data;
+      }, {} as AccountLimitsPatchPayload);
+      this.updateLimits(payload);
+    }
+  }
+
+  cancel() {
+    this.store.getAccount().subscribe((account) => this.updateForm(account));
```
ebrehault, line 101: a `takeUntil` would be needed here, no?

apps/manager-v2/src/app/manage-accounts/account-details/limits/limits.component.ts
```diff
@@ -32,69 +35,152 @@ export class LimitsComponent implements OnInit, OnDestroy {
...
+    };
+    return resetPayload;
+  }
+
+  private updateLimits(payload: AccountLimitsPatchPayload) {
+    this.isSaving = true;
+    this.store
+      .getAccount()
+      .pipe(
+        switchMap((account) =>
```
ebrehault, line 151: takeUntil

ebrehault (approved):

apps/manager-v2/src/app/manage-accounts/account-details/limits/limits.component.ts
mpellerin42: no, `getAccount()` contains a `take(1)`

apps/manager-v2/src/app/manage-accounts/account-details/limits/limits.component.ts
mpellerin42: same, `getAccount()` contains a `take(1)`:
```ts
  getAccount(): Observable<ExtendedAccount> {
    return this.accountDetails.pipe(
      filter((account) => !!account),
      map((account) => account as ExtendedAccount),
      take(1),
    );
  }
```

PR #828, opened 2023-06-13
Confluence connector

ebrehault:


apps/desktop/electron/src/service/nuclia-cloud.ts
```diff
@@ -24,10 +26,15 @@ export class NucliaCloud {
...
   }
 
-  upload(originalId: string, filename: string, data: { buffer?: ArrayBuffer; metadata?: any }): Observable<boolean> {
-    if (data.buffer) {
-      const buffer = data.buffer;
-      const slug = sha256(originalId);
+  upload(
+    originalId: string,
+    filename: string,
+    data: { buffer?: ArrayBuffer; text?: TextField; metadata?: any },
```
mpellerin42, line 32: I don't understand: in `service/models.ts`, you added the following comment:
```
  // we cannot use the TextField from the SDK because we want to keep connectors independant
```
and here in `service/nuclia-cloud.ts` you're actually using `TextField` from the SDK…

apps/desktop/electron/src/service/sources.ts
```diff
@@ -84,12 +85,14 @@ export function getSourceFolders(sourceId: string, query?: string): Observable<S
 function downloadFileOrLink(
   sourceId: string,
   item: SyncItem,
-): Observable<{ type: 'blob' | 'link'; blob?: Blob; link?: any }> {
+): Observable<{ type: 'blob' | 'link' | 'text'; blob?: Blob; link?: any; text?: TextField }> {
```
mpellerin42, line 88: here where `TextField` dependency is coming from `@nuclia/core` and not `SDK-core`. I'm not sure which one is the good one but I feel like we should be consistent, no?

mpellerin42 (changes requested):

apps/desktop/electron/src/service/nuclia-cloud.ts
ebrehault: I meant the source connectors, not the destination ones, so the ones that external people might decide to implement themselves.

mpellerin42 (approved):

PR #830, opened 2023-06-14
[sc-5893] Filter by entities

operramon:


libs/common/src/lib/widgets/widget-generator.component.html
```diff
@@ -63,6 +63,17 @@ <h3>{{ 'stash.widgets.configuration' | translate }}</h3>
             formControlName="filter">
             {{ 'stash.widgets.filter' | translate }}
           </pa-toggle>
+          <div class="filter-type">
+            <pa-select
+              *ngIf="mainFormFeatures.filter"
+              [value]="filters"
+              (valueChange)="onFiltersChange($event)"
+              [label]="'stash.widgets.filter_type' | translate">
+              <pa-option value="both">{{ 'stash.widgets.labels_entities' | translate }}</pa-option>
```
mpellerin42, line 74: I think you should not have an option `both` because it's implying we will always have only two filter options, which is probably not true.
Better than a select, you should probably use a checkbox list (one checkbox by filter type, so for now only two) with all checkboxes selected by default

ebrehault (approved):

libs/search-widget/src/core/search-bar.ts
```diff
@@ -62,10 +61,10 @@ export const setupTriggerSearch = (
                       ...options,
                       show,
                       filters,
-                      inTitleOnly: isTitleOnly(query, labelFilters),
+                      inTitleOnly: !query && allFiltersApplyToResource,
```
mpellerin42, line 64: this doesn't make much sense to me: I remember we search in title only when there is no query, but I don't understand why we also check that all filters are applied

mpellerin42 (changes requested):

libs/common/src/lib/widgets/widget-generator.component.html
operramon: ok, I'll try using checkboxes

libs/search-widget/src/core/search-bar.ts
operramon: Maybe I didn't chose a good name. `allFiltersApplyToResource` is true when all current filters are labels of kind `resource`

mpellerin42 (approved):

PR #831, opened 2023-06-15
Add slider documentation page in sistema doc

mpellerin42:


ebrehault (approved):

apps/sistema-demo/src/app/app.component.ts
```diff
@@ -56,6 +56,7 @@ import {
   SistemaPasswordInputComponent,
   SistemaSpinnerComponent,
 } from './sistema-pages';
+import { SliderPageComponent } from '../../../../libs/pastanaga-angular/projects/demo/src/app/demo/pages/slider-page/slider-page.component';
```
operramon, line 59: A minor thing, the slider has different styles in Firefox and Chrome

operramon (approved):

apps/sistema-demo/src/app/app.component.ts
mpellerin42: @operramon can you open an issue about that on https://github.com/plone/pastanaga-angular/issues please? if you can join screensots would be nice ;)

mpellerin42: I will also add the missing exports in index.ts to fix those import path, but all that has to be done in Pastanaga repo.

PR #836, opened 2023-06-16
[sc-5988] Display warning when getting 529 from chat

operramon:


libs/sdk-core/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 ### Breaking change
 - `setField` method replaces `addField` and `updateField` which were identical.
 
+### Improvements
+- Manage properly 529 status (partial results) on chat endpoint
```
ebrehault, line 7: The 529 error does not mean exactly we have partial results.

Here is the thing:
- The `/chat` endpoint made (internally) a call to `/find` and got a 206 status (which means "Partial results")
- As it makes no sense to generate an answer based on partial results, `/chat` returns a 529, which means "Service is overloaded" (we decided to use 529 here because we know that we get partial results when the system is overloaded)

So we should no mention "partial results" here (enough that's the original problem), the real error is "Service overloaded"

libs/sdk-core/src/lib/db/search/chat.models.ts
```diff
@@ -17,6 +17,7 @@ export namespace Chat {
     id: string;
     sources?: Search.FindResults;
     incomplete?: boolean;
+    partial?: boolean;
```
ebrehault, line 20: rename it `overloaded`

libs/sdk-core/src/lib/db/search/chat.ts
```diff
@@ -46,7 +46,7 @@ export function chat(
             text = text.split(END_OF_STREAM)[0];
           }
         }
-        return { text, sources, incomplete, id };
+        return { text, sources, incomplete, id, partial: status === 529 };
```
ebrehault, line 49: rename it `overloaded`

libs/search-widget/src/core/api.ts
```diff
@@ -91,7 +92,13 @@ export const getAnswer = (query: string, chat?: Chat.Entry[], filters?: string[]
     return acc;
   }, [] as Chat.ContextEntry[]);
 
-  return nucliaApi.knowledgeBox.chat(query, context, [Chat.Features.PARAGRAPHS], { filters });
+  return nucliaApi.knowledgeBox.chat(query, context, [Chat.Features.PARAGRAPHS], { filters }).pipe(
+    tap((res) => {
+      if (res.partial) {
```
ebrehault, line 97: rename it `overloaded`

libs/search-widget/src/core/api.ts
```diff
@@ -91,7 +92,13 @@ export const getAnswer = (query: string, chat?: Chat.Entry[], filters?: string[]
...
+        chatError.set({ type: 'error', status: 529, detail: 'Partial results' });
```
ebrehault, line 98: details must `Service overloaded`

libs/search-widget/src/widgets/search-widget/SearchResults.svelte
```diff
@@ -83,7 +84,7 @@
     {:else if !$pendingResults && $smartResults.length === 0 && !$onlyAnswers}
       <strong>{$_('results.empty')}</strong>
     {:else}
-      {#if $hasPartialResults}
+      {#if $hasPartialResults || $hasChatPartialResults}
```
ebrehault, line 87: here we want a new warning message saying `Sorry, the service is overloaded and could not deliver an answer.`

ebrehault (approved):

libs/sdk-core/CHANGELOG.md
operramon: oh, I'll rework it

ebrehault (approved):

PR #843, opened 2023-06-20
Do not allow dataset import if the kb is monolingual

mpellerin42:


libs/common/src/lib/resources/resource-list/resource-list.component.html
```diff
@@ -145,7 +145,7 @@
     </ng-container>
 
     <div
-      *ngIf="isMainView && emptyKb && !query && !isFiltering && !standalone"
+      *ngIf="isMainView && emptyKb && !query && !isFiltering && !standalone && (isMonoLingual | async) === false"
```
operramon, line 148: should this modification also be added to the `DatasetImportComponent` component ?

operramon (approved):

ebrehault (approved):

libs/common/src/lib/resources/resource-list/resource-list.component.html
mpellerin42: good point, let me check

PR #846, opened 2023-06-21
pdf annotation (using Pawls)

ebrehault:


libs/sdk-core/CHANGELOG.md
```diff
@@ -1,8 +1,10 @@
 # 1.5.0 (unreleased)
 
 ### Breaking changes
+
 - Remove `Search.SmartParagraph` interface as it's not used anymore
 - Rename `Search.SmartResult` interface to `Search.FieldResult` as `Smart` prefix doesn't make sense anymore
+- Add `selections` in field metadata
```
mpellerin42, line 7: doesn't seem like a breaking change to me

mpellerin42 (approved):

libs/sdk-core/CHANGELOG.md
ebrehault: I'll fix it later, I need to deploy

PR #852, opened 2023-06-23
Ericbrehault/sc 5687/integrate posthog on the widget

ebrehault:


libs/search-widget/src/core/tracking.ts
```diff
@@ -0,0 +1,19 @@
+import posthog from 'posthog-js';
+
+let _kbid = '';
+
+export function initTracking(kbid: string) {
+  _kbid = kbid;
+  if (window.location.hostname !== 'localhost') {
+    posthog.init('phc_TAooetKnDVSmvK9WfqtFVUtiEjxpgnPYQ5OaTWUVuY8', { api_host: 'https://eu.posthog.com' });
```
mpellerin42, line 8: isn't it a bit unsafe to have our posthog key like that in an open source project?

mpellerin42 (approved):

PR #861, opened 2023-06-27
[sc-6139] Search results: display one tile by field

mpellerin42:
- Update `getSortedResults` to return one SmartResult entry by field
- Update the state to store the resources coming from the results as a list in `resourceList` so we can concatenate them when loading more
- Don't show duplicated results
- Store resultList in the state
  - Rename search models without `smart` prefix
  - Refactor search state to store results as a list of sorted `FieldResults`
- Fix generative answer bug which was writing in the state a lot while the answer was streaming

### DIFF
--- a/libs/search-widget/src/core/models.ts
+++ b/libs/search-widget/src/core/models.ts
@@ -9,14 +9,27 @@ import type {
   WidgetFeatures,
 } from '@nuclia/core';
 
-export const NO_RESULTS: Search.FindResults = {
+export interface FindResultsAsList extends Omit<Search.FindResults, 'resources'> {
+  resultList: Search.FieldResult[];
+}
+
+const EMPTY_FIND_RESULTS: Omit<Search.FindResults, 'resources'> = {
   type: 'findResults',
-  resources: {} as { [id: string]: Search.FindResource },
   total: 0,
   page_number: 0,
   next_page: false,
   page_size: 0,
   query: '',
+};
+
+export const NO_RESULTS: Search.FindResults = {
+  ...EMPTY_FIND_RESULTS,
+  resources: {},
+};
+
+export const NO_RESULT_LIST: FindResultsAsList = {
+  ...EMPTY_FIND_RESULTS,
+  resultList: [],
 };
 
 export const NO_SUGGESTION_RESULTS: Search.Suggestions = {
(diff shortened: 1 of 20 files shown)


libs/search-widget/src/core/models.ts
```diff
@@ -9,16 +9,29 @@ import type {
...
   type: 'findResults',
-  resources: {} as { [id: string]: Search.FindResource },
   total: 0,
   page_number: 0,
   next_page: false,
   page_size: 0,
   query: '',
 };
 
+export const NO_RESULTS: Search.FindResults = {
```
ebrehault, line 25: not needed anymore

libs/search-widget/src/core/stores/search.store.ts
```diff
@@ -98,10 +98,14 @@ export const searchResults = searchState.writer<
...
   (state) => state.results,
   (state, params) => {
-    CACHE = undefined;
+    const sortedResults = getSortedResults(Object.values(params.results.resources || {}));
     return {
       ...state,
-      results: params.append ? appendResults(state.results, params.results) : params.results,
+      results: {
+        ...params.results,
+        resources: undefined,
```
ebrehault, line 106: not needed

ebrehault (approved):

libs/search-widget/src/core/models.ts
mpellerin42: it is used in Suggestion. That reminds me I forgot to test it

libs/search-widget/src/core/stores/search.store.ts
mpellerin42: If we don't do that, we have the resources duplicated between `resources` and `resultList` property

libs/search-widget/src/core/stores/search.store.ts
```diff
@@ -419,87 +425,54 @@ function appendResults(existingResults: Search.FindResults, newResults: Search.F
...
-    return { field_id: Object.keys(resource.data.files)[0], field_type: FIELD_TYPE.file };
-  } else if (resource.data.links) {
-    return { field_id: Object.keys(resource.data.links)[0], field_type: FIELD_TYPE.link };
-  } else if (resource.data.texts) {
-    return { field_id: Object.keys(resource.data.texts)[0], field_type: FIELD_TYPE.text };
-  } else {
-    return;
+export function getSmartResultUniqueKey(result: Search.SmartResult): string {
+  const key = `${(result.paragraphs || []).reduce((acc, curr) => `${acc}${acc.length > 0 ? '__' : ''}${curr.id}`, '')}`;
+  if (!allKeys.includes(key)) {
```
operramon, line 456: is `allKeys`used ?

PR #865, opened 2023-06-30
[sc-5904] Fix expiration days

operramon:
- Do not display days left when `trial_expiration_date` is `null`
- Display `1 day left` instead of `1 days left`
- Fix calculation to check if an account is expired
- Do not show `Quota exhausted` message


ebrehault (approved):

apps/dashboard/src/app/account/account-status/account-status.component.ts
```diff
@@ -26,29 +25,33 @@ export class AccountStatusComponent {
...
-  canExtendTrial = combineLatest([this.daysLeft, this.trialAccount]).pipe(
+  canExtendTrial = combineLatest([this.daysLeft, this.isTrial]).pipe(
     map(([daysLeft, isTrial]) => isTrial && daysLeft <= 5),
   );
-  trialExpired = combineLatest([this.daysLeft, this.trialAccount]).pipe(
-    map(([daysLeft, isTrial]) => isTrial && daysLeft < 0),
+  trialExpired = combineLatest([this.sdk.currentAccount, this.isTrial]).pipe(
+    map(([account, isTrial]) => {
+      if (!isTrial || !account.trial_expiration_date) return false;
+      const expiration = new Date(`${account.trial_expiration_date}+00:00`);
```
mpellerin42, line 51: I know why you did this trick of adding  `+00:00` at the end of the date in several places, but you should not do that. it's backend's job to send date properly formatted (= with timezone). I already told @albertnadal and @sunbit several times when I was working on the manager, so any time you notice timestamp not formatted as expected, ping them so they fix it on the backend. But **do not** add this kind of kix on the frontend part otherwise we will have bad bugs as soon as the backend will be fixed…

operramon: ok, I'll let them know

PR #885, opened 2023-07-17
[sc-6161] review search results architecture

mpellerin42:


libs/search-widget/src/components/viewer/renderers/VideoRenderer.svelte
```diff
@@ -0,0 +1,64 @@
+<script lang="ts">
+  import type { Observable } from 'rxjs';
+  import { currentThumbnail, getFieldUrl, getFileUrls, getPlayableVideo, isLinkField, playFrom } from "../../../core";
```
ebrehault, line 3: weird, there are double quotes on some of these `import`s…

ebrehault (approved):

libs/search-widget/src/components/viewer/renderers/VideoRenderer.svelte
mpellerin42: it's my IDE which is putting them automatically when updated the imports… I manually put back single quotes, but I might have missed some. I'll try to clean them all up before marking the PR ready for review

libs/search-widget/src/core/api.ts
```diff
@@ -118,23 +120,23 @@ export const searchInResource = (
...
+): Observable<Search.FindResults> => {
   if (!nucliaApi) {
     throw new Error('Nuclia API not initialized');
   }
   options.inTitleOnly = !query && (!options.filters || options.filters.length === 0);
 
   return nucliaApi.knowledgeBox
     .getResourceFromData(resource)
-    .search(query, features, { ...options, ...DEFAULT_SEARCH_OPTIONS })
+    .find(query, features, { ...options, ...DEFAULT_SEARCH_OPTIONS })
```
operramon, line 131: From what I have tested, when searching inside a resource I never get results.

operramon (approved):

PR #909, opened 2023-07-26
Option to hide sources in generated answers

ebrehault:


libs/search-widget/src/components/answer/Answer.svelte
```diff
@@ -26,11 +27,11 @@
 <div class="sw-answer">
   <div class="answer-container">
     <div class="text">{@html text}</div>
-    {#if !isMobile && !hideFeedback}
+    {#if $hideSources && !isMobile && !hideFeedback}
```
operramon, line 30: The feedback button is missing in mobile. Maybe is because of this condition ?

operramon (approved):

libs/search-widget/src/components/answer/Answer.svelte
ebrehault: oh right!
thanks!

PR #927, opened 2023-08-02
mark for check when updating the resources list

ebrehault:


libs/common/src/lib/resources/resource-list/resource-list.component.ts
```diff
@@ -44,6 +44,7 @@ import {
 import { UploadService } from '../../upload/upload.service';
 import { getDesktopAppUrl, getDesktopPlatform, RELEASE_URL } from '../../utils';
 import { NavigationService } from '../../services';
+import { th } from 'date-fns/locale';
```
operramon, line 47: is it used ?

operramon (approved):

operramon (approved):

PR #973, opened 2023-09-01
option to disable fulltext in chat

ebrehault:


libs/search-widget/src/core/api.ts
```diff
@@ -76,6 +76,11 @@ export const initNuclia = (options: NucliaOptions, state: KBStates, widgetOption
   if (widgetOptions.features?.suggestEntities) {
     SUGGEST_MODE.push(Search.SuggestionFeatures.ENTITIES);
   }
+  if (widgetOptions.features?.noBM25forChat) {
+    // no BM25 should be achieved by just removing `paragraphs`, but passing `relations` alone breaks the caht response
```
mpellerin42, line 80:
```suggestion
    // no BM25 should be achieved by just removing `paragraphs`, but passing `relations` alone breaks the chat response
```

mpellerin42 (approved):

libs/sdk-core/src/lib/db/kb/kb.models.ts
```diff
@@ -233,6 +233,7 @@ export interface WidgetFeatures {
   knowledgeGraph?: boolean;
   useSynonyms?: boolean;
   autofilter?: boolean;
+  noBM25forChat?: boolean;
```
operramon, line 236: I think the `chat` endpoint is used to show the first page of results.
Will this option remove bm25 from the search results ?

operramon (approved):

libs/sdk-core/src/lib/db/kb/kb.models.ts
ebrehault: Yes it will.

We are aware this option might be not entirely consistent (and by default it will be false), but Eudald needs it for demo purpose.

PR #995, opened 2023-09-14
[sc-6984] Implement new desktop UX

operramon:


apps/desktop/src/app/sync/sources/sitemap.ts
```diff
@@ -29,15 +28,11 @@ class SitemapImpl implements ISourceConnector {
...
-      this.url = params['url'];
-    }
+    // eslint-disable-next-line no-empty-function
   }
 
   getParametersValues(): ConnectorParameters {
-    return {
-      path: this.url,
-    };
+    return {};
```
ebrehault, line 35: This looks wrong, `getParametersValues` is supposed to return all the config values that will be useful in the sync server, so obviously, we need the sitemap url if we want the server to be able to download it.
Nevertheless, in the `main` branch I see we return `path`, while in the sitemap connector we try to read `sitemap`: ```this.params['sitemap']```

So something is also probably wrong in the current implementation.

apps/desktop/src/app/upload/upload.component.spec.ts
```diff
@@ -21,7 +21,7 @@ import {
 import { MockModule, MockProvider } from 'ng-mocks';
 import { By } from '@angular/platform-browser';
 
-describe('UploadComponent', () => {
+xdescribe('UploadComponent', () => {
```
ebrehault, line 24: typo?

ebrehault (commented):
Looks good. Great work :)

apps/desktop/src/app/upload/upload.component.spec.ts
operramon: I temporarily disabled the test because it needs to be adapted to the new UI. I'll fix it.

ebrehault (approved):

ebrehault (approved):
You also have to increase the version in `package.json`.

PR #1001, opened 2023-09-18
[sc-7111] new section to upload data

mpellerin42:
- Improve home page layout
  - update gaps between blocks
  - make all blocks with consistent look
  - remove download desktop block
  - add upload data block
  - update blocks visibility permissions
- Move connector logos to sistema so any app can access them
- New Upload data page
  - Add source logos in home page
  - Add upload data page in navbar and routing
  - Use same desktop sources component in home page and upload data page
  - Open upload dialog and Desktop app from upload-data page
  - Redirect to resources page after upload
  - Display explicit download and open options for Nuclia desktop


ebrehault (approved):
Looks good to me.
One remark for later:
when https://app.shortcut.com/flaps/story/7012/use-routing-for-pending-resources will be done, we could redirect to the pending list directly (rather than to the normal resource list), it will be more consistent.

PR #1004, opened 2023-09-19
[sc-7253] Error when using different accounts on same connector type

operramon:


mpellerin42 (approved):

apps/desktop/src/app/sync/sources/oauth.ts
```diff
@@ -27,17 +29,17 @@ export class OAuthConnector implements ISourceConnector {
 
   getParametersValues(): ConnectorParameters {
     return {
-      token: localStorage.getItem(this.name + TOKEN),
-      refresh: localStorage.getItem(this.name + REFRESH),
+      token: localStorage.getItem(`${this.name}-${this.id}` + TOKEN),
```
ebrehault, line 32: Refactoring: instead of repeating the same string computation many times, maybe we could have a private method `prefixStorageKey(key: string)` that would do it (I know I come up first with this repetition 😅 ).

ebrehault (approved):

apps/desktop/src/app/sync/sources/oauth.ts
operramon: fixed 👍

PR #1013, opened 2023-09-22
[sc-7012] Resource list refactoring using routing

mpellerin42:
- Resource list refactoring using routing
- Redirect to pending resources after uploading from upload data page

### DIFF
--- a/libs/common/src/lib/resources/resource-list/error-resources-table/error-resources-table.component.ts
+++ b/libs/common/src/lib/resources/resource-list/error-resources-table/error-resources-table.component.ts
@@ -4,7 +4,7 @@ import { EMPTY, expand, map, Observable, reduce, take } from 'rxjs';
 import { IErrorResponse, IResource, KnowledgeBox, Resource, RESOURCE_STATUS, Search } from '@nuclia/core';
 import { switchMap, tap } from 'rxjs/operators';
 import { DEFAULT_PAGE_SIZE, DEFAULT_SORTING } from '../resource-list.model';
-import { UploadService } from '../../../upload';
+import { UploadService } from '../../../upload/upload.service';
 
 @Component({
   selector: 'stf-error-resources-table',
--- a/libs/common/src/lib/resources/resource-list/error-resources-table/error-resources-table.component.ts
+++ b/libs/common/src/lib/resources/resource-list/error-resources-table/error-resources-table.component.ts
@@ -1,9 +1,10 @@
-import { ChangeDetectionStrategy, Component, EventEmitter, Input, Output } from '@angular/core';
+import { ChangeDetectionStrategy, Component, inject } from '@angular/core';
 import { ResourcesTableDirective } from '../resources-table.directive';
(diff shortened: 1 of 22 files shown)


libs/common/src/lib/resources/resource-list/error-resources-table/error-resources-table.component.ts
```diff
@@ -42,24 +38,26 @@ export class ErrorResourcesTableComponent extends ResourcesTableDirective {
   }
 
   override bulkDelete() {
-    const resourcesObs = this.allErrorsSelected ? this.getAllResourcesInError() : of(this.getSelectedResources());
-    resourcesObs.subscribe((resources) => this.deleteResources.emit(resources));
+    const resourcesObs = this.allErrorsSelected ? this.getAllResourcesInError() : this.getSelectedResources();
+    // FIXME
```
ebrehault, line 42: pending fix?

libs/common/src/lib/resources/resource-list/error-resources-table/error-resources-table.component.ts
```diff
@@ -42,24 +38,26 @@ export class ErrorResourcesTableComponent extends ResourcesTableDirective {
...
   override bulkReprocess() {
-    const resourcesObs = this.allErrorsSelected ? this.getAllResourcesInError() : of(this.getSelectedResources());
+    const resourcesObs = this.allErrorsSelected ? this.getAllResourcesInError() : this.getSelectedResources();
     resourcesObs.subscribe((resources) => {
       if (this.allErrorsSelected) {
         this.reprocessAll.emit(resources);
         this.toaster.info('resource.reindex-all-info');
       } else {
-        this.reprocessResources.emit(resources);
+        // FIXME
```
ebrehault, line 53: pending fix?

libs/common/src/lib/resources/resource-list/resource-list.service.ts
```diff
@@ -0,0 +1,273 @@
...
+    this._sort = DEFAULT_SORTING;
+    this._query = '';
+    this._titleOnly = true;
+  }
+
+  get sort(): SortOption {
+    return this._sort;
+  }
+
+  filter(filters: string[]) {
```
ebrehault, line 85: Not a big fan. When a service makes a subscribe in its methods, it is very error prone: when creating a new component we might call it in a subscribe without being aware. The developer has to carefully check how the service method is implemented because the typing will not indicate anything.

2 solutions:
- no subscribe: the method returns an observable and the component is meant to subscribe itself (that's super explicit, typing will indicate that's an observable, but I agree it can be boring and produce more code on the component side)
- decoupling: add in the service a private `_triggerResourceLoad` subject, in constructor, subscribe to the subject to call `loadResource`, and in the `filter` method (and all the other ones subscribing to `loadResource` at the moment), you just do `this._triggerResourceLoad.next()`.
That way the method is entirely safe to be called from wherever (in a subscribe or not).

ebrehault (approved):

operramon (approved):

PR #1025, opened 2023-09-26
[sc-7290] Remove upgrade block for some account types

operramon:


apps/dashboard/src/app/account/account-home/account-home.component.ts
```diff
@@ -61,6 +61,7 @@ export class AccountHomeComponent {
     [StatsType.MEDIA_SECONDS]: (value) => value / 3600,
     [StatsType.TRAIN_SECONDS]: (value) => value / 3600,
   };
+  canUpgrade = this.account$.pipe(map((account) => ['stash-trial', 'stash-starter'].includes(account.type)));
```
ebrehault, line 64: As you use this same list of values in 2 different locations, maybe it would be nice to define a ```export const UPGRADABLE_ACCOUNT_TYPES = ['stash-trial', 'stash-starter'];```
in some place and then. import it in the 2 files.

That way, if we change this list later, it will avoid mistakes.

ebrehault (approved):

apps/dashboard/src/app/account/account-home/account-home.component.ts
operramon: Fixed

PR #1028, opened 2023-09-27
display link to dashboard when no labels

ebrehault:


apps/desktop/src/assets/i18n/en.json
```diff
@@ -33,6 +33,7 @@
 	"upload.add-connector": "Add connector",
 	"upload.connectors": "Connectors",
 	"upload.labels": "Apply labels to resources uploaded through this connector",
+	"upload.no-labels": "If you want to assign labels when uploading files, define some labelsets in your <a href=\"{{url}}\">dashboard</a>.",
```
operramon, line 36: using <a> element, the desktop app is replaced by the labels sets page.
would be better to open a new tab with `openExternal` ?

operramon (approved):

PR #1034, opened 2023-09-29
Fix resource classification loading

mpellerin42:
- add missing mark for check
- add isReady subject to resource classification page
- refactor labelField component to return hasLabel instead of noLabel
- refactor LabelsService replacing `getLabelsByKind` method by two properties: `resourceLabelSets` and `paragraphLabelSets` which emit only when labelSet is not null
- Fix tests (prevent error to be triggered and logged during unit tests)

### DIFF
--- a/libs/common/src/lib/resources/edit-resource/classification/resource-classification.component.html
+++ b/libs/common/src/lib/resources/edit-resource/classification/resource-classification.component.html
@@ -1,21 +1,19 @@
 <div class="main-container">
-  <p *ngIf="noLabels; else labelSelector">
-    {{ 'resource.classification.no-resource-labelset' | translate }}
-    <a [routerLink]="(kbUrl | async) + '/label-sets'">{{ 'resource.classification-page-link' | translate }}</a>
-  </p>
-  <ng-template #labelSelector>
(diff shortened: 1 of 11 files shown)


operramon (approved):

libs/common/src/lib/resources/edit-resource/classification/resource-classification.component.html
```diff
@@ -1,21 +1,19 @@
...
-  <ng-template #labelSelector>
-    <div class="scrollable-area classification-container">
-      <app-label-field
-        [selection]="currentLabels"
-        [disabled]="(isAdminOrContrib | async) === false"
-        (noLabels)="noLabels = $event"
-        (selectionChange)="updateLabels($event)"></app-label-field>
-    </div>
-  </ng-template>
+  <div class="scrollable-area classification-container">
```
operramon, line 2: If this page is refreshed with ⌘-r, then it doesn't display any selected label

operramon: sorry, I didn't test it properly, it works fine

PR #1043, opened 2023-10-03
[sc-7329] implement navigation in resource preview

mpellerin42:
- Setup navigation between resources in edit flow
- Fix select first field to select oldest field when possible
- Prevent `/catalog` requests sent by resource list (to load the list, the filters and the status counts) to be made in parallel: we need to use the same shard for all those requests


libs/common/src/lib/resources/edit-resource/add-field/add-field.component.ts
```diff
@@ -53,7 +55,14 @@ export class AddFieldComponent implements OnInit, OnDestroy {
...
   };
 
-  constructor(private editResource: EditResourceService, private cdr: ChangeDetectorRef) {}
+  constructor(
+    private editResource: EditResourceService,
+    private cdr: ChangeDetectorRef,
+    private route: ActivatedRoute,
+    private resourceNavigation: ResourceNavigationService,
+  ) {
+    this.resourceNavigation.currentRoute = this.route;
```
ebrehault, line 64: Injecting a service in a component in order to pass the service object itself to another service seems weird.

What about:
- injecting `ActivatedRoute` in `ResourceNavigationService` only
- add a method in `ResourceNavigationService`:
```ts
storeCurrentRoute() {
    this._snapshot = this.route.snapshot;
}```

And the 2 components would just call this method.

ebrehault (approved):

libs/common/src/lib/resources/edit-resource/add-field/add-field.component.ts
mpellerin42: route ([ActivatedRoute](https://angular.io/api/router/ActivatedRoute)) is not a service, it is tight to a component:
> Provides access to information about a route associated with a component that is loaded in an outlet.

so we can't inject it directly in a service (I tried it like that first, but the route was wrong)

PR #1044, opened 2023-10-03
[sc-7355] Translate apps and widgets in german and portuguese

mpellerin42:
- Add german and portuguese translations on common lib
- user lib
- widget and desktop


libs/common/src/assets/i18n/en.json
```diff
@@ -614,14 +614,11 @@
 	"resource.title": "Resource title",
 	"resource.tooltip.add-field": "Add a field",
 	"resource.tooltip.annotation": "Annotation",
-	"resource.tooltip.back-to-resources": "Back to resource list",
```
operramon, line 617: should these texts be deleted ?

libs/common/src/assets/i18n/de.json
```diff
@@ -0,0 +1,892 @@
+{
+	"account.account_name_invalid": "Ungültiger Kontoname",
+	"account.accumulated_consumption": "Kumulierter Tagesverbrauch nach Funktion",
+	"account.active_config": "Aktive Konfiguration",
+	"account.add_stash": "Wissensbox hinzufügen",
+	"account.add_user": "Nutzer hinzufügen",
+	"account.admin.make_admin": "Administrator erstellen",
+	"account.admin.warning": "Wenn Sie diesem Benutzer die Administratorrolle zuweisen, kann er auf alle Optionen zugreifen und diese bearbeiten.",
+	"account.billing": "Zahlung & Abonnement",
+	"account.chart_error_media_seconds": "Die Daten zu „Medienstunden“ konnten nicht abgerufen werden",
```
DomFleischmann, line 10: For some reason this was translated to "media hours" instead of "media seconds" should be something like "Mediensekunden" (please double check).

libs/common/src/assets/i18n/de.json
```diff
@@ -0,0 +1,892 @@
...
+	"account.chart_error_processing_status": "Die Daten „zur Verarbeitung ausstehende Ressourcen“ konnten nicht abgerufen werden",
+	"account.chart_error_searches": "„Suchdaten“ können nicht abgerufen werden",
+	"account.chart_error_train_seconds": "Die Daten zu „Zugstunden“ können nicht abgerufen werden",
```
DomFleischmann, line 13: This one translated "train seconds" to "train hours" but train as in the transportation method 🚂 . It should probably use the english word training like "Trainingsekunden".

libs/common/src/assets/i18n/de.json
```diff
@@ -0,0 +1,892 @@
...
+	"account.chart_media_seconds": "Medienzeiten",
```
DomFleischmann, line 14: This was translated to "media times" instead of "media seconds".

libs/common/src/assets/i18n/de.json
```diff
@@ -0,0 +1,892 @@
...
+	"account.chart_number_resources": "Anzahl der Ressourcen",
+	"account.chart_range_label": "Letzte {{hours}}",
+	"account.chart_searches": "Suchen",
+	"account.chart_train_seconds": "Trainingsstunden",
```
DomFleischmann, line 18: Again hours, instead of seconds. But it translated training correctly.

libs/common/src/assets/i18n/de.json
```diff
@@ -0,0 +1,892 @@
...
+	"account.code_saved": "Code eingegeben",
+	"account.config": "Kontoeinstellungen",
+	"account.consuming": "Konsumierend",
```
DomFleischmann, line 21: I think this was translated to the wrong verbal time, as it should be something ongoing.

libs/common/src/assets/i18n/de.json
```diff
@@ -0,0 +1,892 @@
...
+	"account.create_new_db": "Erstellen Sie ein neues Nuclia DB-Konto",
+	"account.create_nua_key": "Erstellen Sie einen neuen Nuclia Understanding-API-Schlüssel",
```
DomFleischmann, line 23: Not sure if the `-` should be like this here.

libs/common/src/assets/i18n/de.json
```diff
@@ -0,0 +1,892 @@
...
+	"account.expired": "Abgelaufen",
+	"account.extend_trial": "Verlängern Sie die kostenlose Testversion",
+	"account.free_trial": "Kostenlose Testphase",
+	"account.free_trial_ended.description": "Die 15-tägige Testversion ist abgelaufen und Ihr Konto wurde gesperrt. Sie können entweder einen unserer Pläne abonnieren oder uns für eine Verlängerung der Testversion kontaktieren.",
+	"account.free_trial_ended.title": "Oh! Ihre kostenlose Testversion ist beendet",
+	"account.full_name": "Name und Nachname",
+	"account.generate_key": "Generieren Sie einen neuen Schlüssel",
+	"account.home": "Benutzerkonto Übersicht",
+	"account.invited_user": "Wir haben eine Einladungs-E-Mail an {{user}} gesendet, um diesem Konto beizutreten",
+	"account.kb-description": "Eine Knowledge Box ist eine Sammlung von Inhalten, nach denen Nuclia sucht. Einige Suchmaschinen nennen es „Index“ oder „Korpus“. In den meisten Fällen benötigen Sie für Ihre Bewerbung nur eine Knowledge Box.",
```
DomFleischmann, line 54: Probably want to be consistent with either `Knowledge Box` or `Wissensbox`

DomFleischmann (commented):
Here are some initial comments on the german translation, only checked the first 100 lines. Will check more during the day.

libs/common/src/assets/i18n/de.json
mpellerin42: actually the english translation is: `Unable to fetch "Media hours" data`.
I'm not sure why the translation key says `seconds`…

libs/common/src/assets/i18n/de.json
mpellerin42: same here, english is `Unable to fetch "Train hours" data`. I guess a better english translation would be `Unable to fetch "Training hours" data`, so the translator then doesn't think we're talking about trains (we had the same problem with french)

libs/common/src/assets/i18n/de.json
mpellerin42: English is: `Media hours` (I guess we changed the way to talk about those charts at some point, because I agree the translation keys doesn't make much sense here)

libs/common/src/assets/i18n/de.json
mpellerin42: probably because the english translation is `Training hours` :)

libs/common/src/assets/i18n/de.json
mpellerin42: Ok, let's forget about this one, I was trying to find it in our app to understand the context and realised it's not in use anymore…

libs/common/src/assets/i18n/de.json
mpellerin42: English is: `Create new Nuclia understanding API key `.
This one is hard to translate, I know the french doesn't make much sense for instance but I have no idea how to update it… If you have a suggestion for the german, let us know.

libs/common/src/assets/i18n/de.json
mpellerin42: We want to be consistent for sure. Which one do you think would be the best option? I'll then make sure to use the same one everywhere.

libs/common/src/assets/i18n/de.json
mpellerin42: removed.

libs/common/src/assets/i18n/de.json
DomFleischmann: I´m not entirely sure. German speakers are pretty used to English words, but at the same time Wissensbox is pretty short and catchy. I think in the end it really depends if Knowledge Box is some kind of brand thing for us.

libs/common/src/assets/i18n/de.json
```diff
@@ -0,0 +1,895 @@
...
+	"billing.hour": "Stunde",
+	"billing.hours": "Std",
+	"billing.ingestion_cost.description": "Die Berechnung hängt von der Anzahl der Absätze der hochgeladenen Dokumente ab. Mit dem Budgetrechner können Sie ein Monatsbudget basierend auf Ihren spezifischen Anforderungen erstellen.",
+	"billing.ingestion_cost.title": "Einnahmekosten",
+	"billing.invalid_address": "Die angegebene Adresse ist ungültig",
+	"billing.invalid_card": "Wir konnten Ihre Karte nicht validieren.",
+	"billing.invalid_vat": "Bitte fügen Sie den Ländercode hinzu. z.B.: ESB000000N",
+	"billing.month": "Monat",
+	"billing.monthly_estimation.description": "Die angegebenen geschätzten Kosten basieren auf den definierten Werten und können sich ändern, wenn diese Werte geändert werden.",
+	"billing.monthly_estimation.title": "Es wird geschätzt, dass übermäßiger Verbrauch Kosten verursacht",
```
DomFleischmann, line 180: This last verb does not make sense.

libs/common/src/assets/i18n/de.json
```diff
@@ -0,0 +1,895 @@
...
+	"billing.reactivate": "Reaktivieren",
+	"billing.renew-subscription": "Abo verlängern",
+	"billing.review.confirm_subscription": "Bestätige dein Abonnement",
+	"billing.review.cost": "Monatliches {{cost}} -Abonnement + Verbrauchskosten mit einer monatlichen Budgetbeschränkung von {{budget}}",
+	"billing.review.subscribe_to": "Abonnieren von {{account}}",
+	"billing.review.title": "Überprüfen Sie Ihr Abonnement",
+	"billing.review_confirm": "Überprüfen und bestätigen",
+	"billing.save_budget": "Als Budgetlimit speichern",
+	"billing.save_continue": "Speichern fortsetzen",
+	"billing.self_hosted": "<a href=\"{{ url }}\">Kontaktieren Sie uns</a> , wenn Sie möchten, dass Ihre Daten mit NucliaDB selbst gehostet werden, um alle Zahlungsdetails zu kennen.",
```
DomFleischmann, line 225: self hosted wasn't translated correctly.

libs/common/src/assets/i18n/de.json
mpellerin42: I agree, and actually we should be consistent in all languages: I don't know if this was translated in spanish/catalan or not. I'll double check with Product team.

libs/common/src/assets/i18n/de.json
```diff
@@ -0,0 +1,895 @@
...
+	"empty.title": "Keine Ergebnisse.",
+	"entities.law": "Rechtliches",
+	"entities.quantity": "Mengen und Maße",
+	"error.creating-kb": "Beim Erstellen der Wissensbox ist ein Fehler aufgetreten",
+	"error.deleting-resource": "Das Löschen der Ressource ist fehlgeschlagen",
+	"error.deleting-resources": "Das Löschen von {{count}} Ressourcen ist fehlgeschlagen",
+	"error.reprocessing-resource": "Die erneute Verarbeitung der Ressource ist fehlgeschlagen",
+	"error.reprocessing-resources": "Die erneute Verarbeitung ist für {{count}} Ressourcen fehlgeschlagen",
+	"farewell.see_again": "Wir hoffen Sie wieder zu sehen.",
+	"farewell.thanks_trust": "Danke für dein Vertrauen!",
```
DomFleischmann, line 280: There is some inconsistency, here it is using informal you while usually it is using formal you.

DomFleischmann (commented):
Some more comments, made it until line 300 this time.

ebrehault (approved):

operramon (approved):

PR #1050, opened 2023-10-06
[sc-7419] Support local NucliaDB as destination in Desktop app

operramon:


libs/common/src/lib/label/label-field/label-field.component.ts
```diff
@@ -25,15 +25,24 @@ export class LabelFieldComponent {
...
 
-  labelSets$ = this._kind.pipe(
-    switchMap((kind) =>
-      kind === LabelSetKind.RESOURCES ? this.labelsService.resourceLabelSets : this.labelsService.paragraphLabelSets,
+  labelSets$ = merge(
+    this._labelSets,
+    this._kind.pipe(
+      switchMap((kind) =>
+        kind === LabelSetKind.RESOURCES ? this.labelsService.resourceLabelSets : this.labelsService.paragraphLabelSets,
+      ),
```
mpellerin42, line 45: I don't understand well why you do a merge here… if you provide the labelSets as an input, do you still need to get the labelsets by kind from `labelsService`?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "nuclia",
-  "version": "2.6.20",
```
ebrehault, line 3: maybe it deserves to be 2.7.0 as that's a new feature

ebrehault (approved):

libs/common/src/lib/label/label-field/label-field.component.ts
operramon: `kind` and `labelSets` inputs are exclusive, only one of the should be provided.
Would you replace it by something like this ?
```
  ngOnInit() {
    if (labelsets input is provided) {
      this.labelsets$ = this._labelSets...
    }
    else {
      this.labelSets$ = this._kind...
    }
  }
```

mpellerin42: OK, I think the fact those inputs are exclusive should be documented in the code then (a simple comment above the inputs would work)

operramon: Fixed

mpellerin42 (approved):

PR #1054, opened 2023-10-09
allow to delete kb from contributor

ebrehault:


libs/common/common.babel
```diff
@@ -17688,6 +17688,29 @@
...
+											<translation>
+												<language>es-ES</language>
+												<approved>false</approved>
+											</translation>
+											<translation>
+												<language>fr-FR</language>
+												<approved>false</approved>
+											</translation>
+										</translations>
+									</concept_node>
```
mpellerin42, line 17713: german and portuguese are missing, this will cause the same problem I had earlier, it would be better for you to rebase and add the missing translations

libs/common/src/lib/select-account-kb/select-kb/select-kb.component.scss
```diff
@@ -74,9 +74,19 @@
       color: $color-text-menu-option-disabled;
       pointer-events: none;
     }
+    .delete-btn {
+      visibility: hidden;
```
mpellerin42, line 78: better to use `opacity: 0`, so you can add a `transition: opacity $transition-hint` as well

mpellerin42 (approved):

PR #1055, opened 2023-10-09
[sc-7470]: add book a demo button in header

mpellerin42:
Also update left and right spaces on the layout to match the latest designs.


libs/common/src/lib/topbar/topbar.component.ts
```diff
@@ -39,4 +40,8 @@ export class TopbarComponent {
       this.router.navigate([url]);
     });
   }
+
+  bookDemo() {
+    window.open('https://calendly.com/nuclia/30min?back=1&month=2023-10', 'blank', 'noreferrer');
```
ebrehault, line 45: does that mean it works only for October 2023?

ebrehault (approved):

libs/common/src/lib/topbar/topbar.component.ts
mpellerin42: no, I tested the link: when passing dates in the past it automatically put the current month, so we don't need to take care of it. But if you think it's safer, I can add a system to update the link according to the current date.

PR #1064, opened 2023-10-13
do not break if kb cannot be accessed

ebrehault:
the fix is not as ambitious as the story, it does not show any small icon next to the failing connector in the UI, but at least the server does not crash and the user can see a proper error message in the activity log


mpellerin42 (approved):
minimal fix indeed XD

PR #1076, opened 2023-10-18
[sc-7575] Block all features on contributor app if no NUA key is provided

operramon:


libs/common/src/lib/navbar/navbar.component.html
```diff
@@ -23,7 +23,8 @@
             [routerLink]="kbUrl + '/upload'"
             class="nav-link"
             routerLinkActive="active"
-            [routerLinkActiveOptions]="{ exact: true }">
+            [routerLinkActiveOptions]="{ exact: true }"
+            [class.disabled]="!(isAdminOrContrib | async) || (invalidKey | async)">
```
mpellerin42, line 27: I don't see how a simple css class can prevent user to click on the link and navigate

operramon: mm.. It prevents clicking it with `pointer-events: none`, but I see it doesn't disable keyword interaction

mpellerin42: ok, it's a bit corner case then, but you can make it fully disabled by adding a condition in `routerLink` as well, like 

```ts
[routerLink]="(isAdminOrContrib | async) === true && (invalidKey | async) === false ?  kbUrl + '/upload' : null"
```

operramon: Fixed

ebrehault (approved):

PR #1081, opened 2023-10-20
Fix [sc-7350]: replace app-label-field by labels-expander in desktop settings

mpellerin42:
- move `LabelsExpanderComponent` and `classification.helpers.ts` from `common` to `sistema`
- add `getSelectionFromClassification` function in helpers
- update `LabelsExpanderComponent` to be standalone
- use `LabelsExpanderComponent` instead of `app-label-field` in desktop settings component


apps/desktop/src/app/upload/settings/settings.component.html
```diff
@@ -137,10 +137,24 @@ <h2 class="unit-test">{{ 'upload.source.settings' | translate }}</h2>
...
+          {{ 'upload.labels' | translate }}
+        </label>
+        <div class="labels-container">
+          <nsi-labels-expander
+            id="label-expander"
+            [labelSets]="labelSets | async"
+            [currentSelection]="currentSelection"
+            (updateSelection)="updateLabelSelection($event)"></nsi-labels-expander>
+          <div>
+            <app-label-list
```
operramon, line 152: I created a connector with some labels and works fine.
Later if go back to the connector settings page, the label list is displayed empty

operramon (approved):

apps/desktop/src/app/upload/settings/settings.component.html
mpellerin42: @operramon fixed

PR #1085, opened 2023-10-25
Hixe account type button when not relevant

ebrehault:


libs/common/src/lib/topbar/topbar.component.ts
```diff
@@ -15,8 +15,18 @@ export class TopbarComponent {
   userInfo = this.userService.userInfo;
   kb = this.sdk.currentKb;
   isStage = location.hostname === 'stashify.cloud';
-  accountType = this.sdk.currentAccount.pipe(map((account) => account.type));
-  isAccountManager = this.sdk.currentAccount.pipe(map((account) => account.can_manage_account));
+  private _accountType = this.sdk.currentAccount.pipe(shareReplay());
```
mpellerin42, line 18: this is not correctly named, it's an account, not an account type

mpellerin42 (approved):

PR #1100, opened 2023-11-06
[sc-7748] Don't load all KBs from all accounts after login

mpellerin42:


ebrehault (approved):
It looks a bit complex.
The service exposes
- `loadKbs` which do load the kbs and return them,
- `kbs` which does not load them, just return them
- it does not manage standalone kbs, so `SelectKbComponent` has to provide its own implementation for `loadKbs` and `kbs` (and just default to the ones from the service when not standalone)

Why not just have everything implemented once in the service?

libs/common/src/lib/select-account-kb/select-account-kb.service.ts
```diff
@@ -12,18 +14,54 @@ export class SelectAccountKbService {
...
+              id: kb.uuid,
+              slug: kb.uuid,
+              zone: 'local',
+              title: kb.slug,
+              role_on_kb: 'SOWNER' as KBRoles,
+            })),
+          ),
+          catchError((error) => {
+            this.toast.error(
+              'We cannot load your knowledge box, please check NucliaDB docker image is running and try again.',
```
ebrehault, line 46: Docker is not the only way to run NucliaDB (it can be a Python install for example, and in the future it might be a service in a Linux distibution), so let's be more neutral:
```
We cannot load your knowledge box, please check NucliaDB server is running and try again.```

ebrehault (approved):

operramon (approved):

PR #1111, opened 2023-11-10
[sc-7737] implement new regional KB endpoints in Dashboard

mpellerin42:
- Implement new regional endpoints in Dashboard
  - Update the SDK core for supporting regional endpoints related to knowledge boxes
  - Add accountId in Nuclia options and set it whenever the account is set
  - Set kb id and kb zone in Nuclia options whenever the KB is selected
  - Add `zone` in dashboard route
    - rename `stash` param into `kb` in dashboard and contributor route
  - Add missing translations for (un)publishing KB confirmation dialog title
  - Fix refresh KB list after updating KB
- Clean up core
  - remove unused pipes from stf-pipes.module
  - replace `DeprecatedApiService` by `SDKService` in all services still using it
  - delete `DeprecatedApiService`
  - remove unused `successResult` method from tracking service (allowing to get rid of unused `DeprecatedApiService` injected dependency)
- Move user management from `core` lib to `sdk-core`
  - account users management methods in `db`
  - KB users management method in `kb`
  - remove `users.service.ts` and `users.model.ts` from `core` lib


apps/dashboard/src/app/account/account-kbs/account-kbs.component.ts
```diff
@@ -125,7 +115,6 @@ export class AccountKbsComponent implements OnInit, OnDestroy {
           this.setLoading(true);
           return new WritableKnowledgeBox(this.sdk.nuclia, this.account!.slug, kb).modify({ state });
         }),
-        switchMap(() => this.updateKbs()),
```
operramon, line 128: It seems that when a kb is published/unpublished, the button text doesn't change

mpellerin42:
Moving back to draft as I realised the contributor app is broken when enabling regional endpoints (while it shouldn't be impacted). I'm investigating what's going on + I'll fix @operramon ’s comments

ebrehault (commented):
Looks good

operramon (approved):

PR #1124, opened 2023-11-13
[sc-7600] Update subscriptions page

operramon:
Improve subscription table:
- Display unit price instead of the threshold when the threshold is 0
- Display the pricing for enterprise tier

Improve calculator:
- Add predict and storage parameters
- Simplify search cost calculations
- Disable search and storage sliders when on-prem mode is selected


libs/common/src/assets/i18n/en.json
```diff
@@ -173,20 +174,21 @@
 	"billing.invalid_vat": "Please add the country code. e.g: ESB000000N",
 	"billing.month": "Month",
 	"billing.monthly_estimation.description": "The estimated cost provided is based on the defined values and is subject to change if these values are modified.",
-	"billing.monthly_estimation.title": "Overconsumption is estimated to cost",
+	"billing.monthly_estimation.title": "Monthly consumption is estimated to cost",
 	"billing.my-subscription": "My subscription",
 	"billing.next-payment": "Next payment",
 	"billing.not_company": "I'm not a company",
+	"billing.on_premise_cost": "With <b>NucliaDB On-prem</b> does not carry any extra cost.",
```
mpellerin42, line 181: as far as I know, `<b>` is kinda deprecated (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b). you should use `<strong>` instead

mpellerin42 (approved):

libs/common/src/assets/i18n/en.json
operramon: I didn't remember, fixed

mpellerin42 (approved):

PR #1139, opened 2023-11-17
[sc-7865] Support preselected filters in widgets

mpellerin42:
- [sc-7875] Use filter alias syntax in widget
- [sc-7865] Support preselected filters in widgets
-  Update de widget generator accordingly


libs/search-widget/src/core/stores/effects.ts
operramon, line 1: Found a small issue related to the url params.
Way to reproduce:
- Select some filters (not pre-filters) and make a search
- Refresh the browser
- Seems like filters are not recovered from the url

operramon (approved):

apps/search-widget-demo/src/App.svelte
```diff
@@ -125,7 +125,7 @@
         lang="en"
         placeholder="Search"
         {prompt}
-        filters="labels,created"
+        preselected_filters="/classification.labels/artist/Mademoiselle K"
```
ebrehault, line 128: Maybe better to do it with a variable like `prompt` so it is easy to comment/comment out.

ebrehault (approved):

PR #1145, opened 2023-11-20
[sc-7880] Apply filters on selection

operramon:
- When search filters change, they are applied immediately.
- In the case of the filters dropdown, filters are applied as soon as the dropdown is closed.

### DIFF
--- a/libs/search-widget/src/components/search-input/SearchInput.svelte
+++ b/libs/search-widget/src/components/search-input/SearchInput.svelte
@@ -232,41 +232,60 @@
   </div>
 
   {#if $filters.length > 0}
-    <div class="filters-container" bind:this={filterContainerElement}>
+    <div
+      class="filters-container"
+      bind:this={filterContainerElement}>
       {#each $filters as filter (filter.key)}
         {#if filter.type === 'creation-start'}
           <Chip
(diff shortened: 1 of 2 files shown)


libs/search-widget/src/components/search-input/SearchInput.svelte
```diff
@@ -232,41 +232,60 @@
...
+      class="filters-container"
+      bind:this={filterContainerElement}>
       {#each $filters as filter (filter.key)}
         {#if filter.type === 'creation-start'}
           <Chip
             removable
             color={entitiesDefaultColor}
-            on:remove={() => creationStart.set(undefined)}>
-            {$_('input.from')} {filter.value}
+            on:remove={() => {
```
ebrehault, line 243: Instead of putting JS code directly in all these `on:remove` events, maybe we should manage everything in the `search()` function, allowing `filter` as optional parameter.

That would be cleaner I think.

ebrehault (approved):

libs/search-widget/src/components/search-input/SearchInput.svelte
operramon: Fixed

PR #1146, opened 2023-11-20
show required fields explicitly on sign up

ebrehault:


mpellerin42 (changes requested):
Not the right way to do this. Simply adding `(required)` in corresponding labels, or as `help` message on each field would be way better

mpellerin42 (approved):

PR #1149, opened 2023-11-21
[sc-7863] Allow labels with underscore in labels expender component

operramon:


ebrehault (approved):

libs/sistema/src/lib/labels-expander/labels-expander.component.html
```diff
@@ -4,7 +4,7 @@
     <div class="label-selector">
       <pa-checkbox
         *ngFor="let label of labelset.value.labels"
-        [value]="currentSelection[labelset.key + '_' + label.title]"
+        [value]="currentSelection[getSelectionKey(labelset.key, label.title)]"
```
mpellerin42, line 7: it would be better to either create a small pipe for this, or to use directly the SEPARATOR variable you created. regular functions are called a lot during the change detection cycles, and performances can be impacted otherwise.

operramon: Fixed

PR #1151, opened 2023-11-22
improve prompt

ebrehault:


operramon (approved):

libs/sdk-core/src/lib/db/search/chat.ts
```diff
@@ -103,44 +103,54 @@ export function chat(
...
-                return {
-                  type: 'answer',
-                  text: answer,
-                  id: 'N/A',
-                  sources: { ...results, resources: matchingResources },
-                } as Chat.Answer;
-              }),
-            );
+          const completeQuery = `${query}. At the end of the answer, add all the corresponding nInternal source references as a comma separated list formatted like this: INTERNAL SOURCES: 1, 2, 13\n\nDO NOT mention the internal source references when phrasing the answer itself.`;
+          console.log('completeQuery', completeQuery);
```
operramon, line 126: are these console.log added on purpose ?

ebrehault: oh no…
Thanks :)

PR #1169, opened 2023-12-05
[sc-7722] new onboarding + getting started flow 

mpellerin42:
- New onboarding flow
  - New wording on sign-up and first questions
  - Refactor onboarding in several steps
  - Update zone service to optionally return zones not available yet
  - Don't display "create account" and "manage account" options at the same time in the user menu
  - Update sistema default border radius to be 0 as in figma (impacting mainly checkboxes and popups)
- New getting started modal
  - New `getting-started` modal with one component by step (intro, upload, processing)
  - Create `upload.utils.ts` with some upload utility methods like `getFilesByType`, `getFilesGroupedByType`
  - Refactor uploadService:
     - we can now use its uploadFiles method without triggering any toast
     - new method `checkFileTypesAndConfirm` for warnings about spreadsheet and archive uploads
  - Move progress-bar component to sistema
  - Update all project target to esnext, so we can use the newish `toSpliced` array method (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSpliced) and its siblings (toSorted, toReversed…) (I finally not used it in this PR)


apps/dashboard/src/app/onboarding/getting-started/getting-started.component.ts
```diff
@@ -0,0 +1,237 @@
...
+                    return item;
+                  });
+
+                  // if everything is processed or failed, we stop polling processing status and enable next button
+                  if (this.itemsToUpload.every((item) => item.processed || item.uploadFailed)) {
+                    this.allProcessed.next();
+                    this.nextDisabled = false;
+                    this.totalEstimatedTime = 0;
+                  } else {
+                    this.totalEstimatedTime = this.itemsToUpload.reduce((total, item) => {
```
operramon, line 183: would it be more accurate to calculate it by getting the item with maximum processing time, instead of adding all the times ?

libs/user/src/lib/onboarding/onboarding.service.ts
```diff
@@ -1,132 +1,191 @@
...
+      accountCreated: false,
+      kbCreated: false,
+      creationFailed: false,
+      datasetImported: false,
+    });
+    this._kbCreationFailureCount = 0;
+
+    const accountSlugRequested = STFUtils.generateSlug(configuration.company);
+    this.tracking.logEvent('account_creation_submitted');
+    this.getAvailableAccountSlug(accountSlugRequested)
```
operramon, line 63: I get a slug like `nuclia____4`, are these underscores expected ?

apps/dashboard/src/app/onboarding/getting-started/getting-started.component.ts
mpellerin42: I'm not sure what you mean by "to calculate it by getting the item with maximum processing time". But in any case, the current naive approach is temporary, the time for the backend to implement a new endpoint providing us better estimations based on uploaded resources type and size.

PR #1195, opened 2023-12-13
use sync agent

ebrehault:
that's temporary integration

### DIFF
--- a/apps/dashboard/src/app/app.module.ts
+++ b/apps/dashboard/src/app/app.module.ts
@@ -38,6 +38,8 @@ import {
 import { KnowledgeBoxModule } from './knowledge-box';
 import { InviteModule } from './invite/invite.module';
 import { MultiTranslateHttpLoader } from 'ngx-translate-multi-http-loader';
+import { SyncService } from '@nuclia/sync';
+import { NewSyncService } from '../../../../libs/sync/src/lib/sync/new-sync.service';
 
 registerLocaleData(localeEn);
 registerLocaleData(localeEs);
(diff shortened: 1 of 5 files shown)


libs/sync/src/lib/sync/new-sync.service.ts
```diff
@@ -0,0 +1,476 @@
...
+  );
+  private _basePath = new BehaviorSubject<string>('/');
+  basePath = this._basePath.asObservable();
+
+  constructor(
+    private sdk: SDKService,
+    private http: HttpClient,
+    private config: BackendConfigurationService,
+  ) {
+    console.log('NewSyncService');
```
operramon, line 140: console.log

apps/dashboard/src/app/app.module.ts
```diff
@@ -38,6 +38,8 @@ import {
 import { KnowledgeBoxModule } from './knowledge-box';
 import { InviteModule } from './invite/invite.module';
 import { MultiTranslateHttpLoader } from 'ngx-translate-multi-http-loader';
+import { SyncService } from '@nuclia/sync';
+import { NewSyncService } from '../../../../libs/sync/src/lib/sync/new-sync.service';
```
operramon, line 42: is wrong import ?

operramon (approved):

libs/sync/src/lib/sync/new-sync.service.ts
ebrehault: yes I keep it to make sure I am using the temporary new service

PR #1211, opened 2023-12-19
Put account creation under feature flag

mpellerin42:


libs/user/src/assets/i18n/en.json
```diff
@@ -29,6 +29,7 @@
 	"login.google-workspace": "Google Workspace",
 	"login.have_account": "Don’t have an account?",
 	"login.login": "Log in",
+	"login.onboarding-disabled": "Account creation is temporarily disabled as part of a maintenance operation. She will be restored quickly.",
```
ebrehault, line 32: `It will be` not `she`
You should have use english as base language before translating

ebrehault (changes requested):

ebrehault (approved):

PR #1213, opened 2023-12-19
Ericbrehault/sc 8156/final migration to new regional endpoints

ebrehault:


libs/sdk-core/src/lib/test.utils.spec.ts
```diff
@@ -13,6 +13,22 @@ export const mockFetch = (response: any) => {
   ) as jest.Mock;
 };
 
+export const multiMockFetch = (responses: { [url: string]: any }) => {
+  global.fetch = jest.fn((url: string) => {
+    console.log('url', url);
```
mpellerin42, line 18: cleanup `console.log`

operramon (approved):

PR #1232, opened 2023-12-28
[sc-8131] New Label sets page layout

mpellerin42:
Waiting for design approval on new label set colors.

Once merged, we should also merge https://github.com/nuclia/e2e/pull/43


libs/common/src/lib/label/label-sets/label-set/label-set.component.ts
```diff
@@ -33,6 +33,14 @@ export class LabelSetComponent implements OnDestroy {
     title: new FormControl<string>('', { validators: [Validators.required, Sluggable()], nonNullable: true }),
     kind: new FormControl<LabelSetKind | undefined>(undefined, { validators: [Validators.required] }),
     exclusive: new FormControl<boolean>(false),
+    labels: new FormControl<string>('', {
+      nonNullable: true,
+      updateOn: 'blur',
+      validators: [
```
operramon, line 39: Regarding label names, the character `/` and `{`,`}` weren't allowed in the previous labels page.
https://github.com/nuclia/frontend/blob/main/libs/common/src/lib/label/label-sets/label-set/label/label.component.html#L14
Should this restriction be kept ?
I guess `/` was restricted because label filters include this character.
I don't know why curly braces were restricted.

operramon (approved):

libs/common/src/lib/label/label-sets/label-set/label-set.component.ts
mpellerin42: you are right, I forgot about that! I'll fix it before merging, thanks for catching this.

PR #1242, opened 2024-01-05
[sc-8194] Fix cache param on translations

operramon:


libs/core/src/lib/config/app.init.service.ts
```diff
@@ -40,9 +40,10 @@ export class AppInitService {
   // This is the method you want to call at bootstrap
   // Important: It should return a Promise
   public init(staticEnv: StaticEnvironmentConfiguration) {
+    const timestamp = Math.floor(Date.now() / 1000);
     return firstValueFrom(
       from(
-        fetch('assets/deployment/app-config.json').then(function (response) {
+        fetch(`assets/deployment/app-config.json?t=${timestamp}`).then(function (response) {
```
mpellerin42, line 46: why don't you keep it simple like calling directly `Date.now()` instead of computing a timestamp based on it?
```suggestion
        fetch(`assets/deployment/app-config.json?t=${Date.now()}`).then(function (response) {
```

operramon: ok 👍

mpellerin42 (approved):

PR #1246, opened 2024-01-08
[sc-3834] Allow to annotate using keyboard

operramon:


ebrehault (approved):

libs/common/src/lib/resources/edit-resource/annotation/paragraph-annotation/paragraph-annotation.component.ts
```diff
@@ -300,4 +301,36 @@ export class ParagraphAnnotationComponent extends SelectFirstFieldDirective impl
       ),
     );
   }
+
+  @HostListener('window:keyup', ['$event'])
+  private onKeyUp(event: KeyboardEvent) {
+    if (document.activeElement !== document.body) return;
+    forkJoin([this.paragraphs.pipe(take(1)), this.isAdminOrContrib.pipe(take(1))]).subscribe(
```
mpellerin42, line 308: this seems a bit heavy to me as you will do this call on each keyup event, while you need it only on Enter, Delete and Escape. It would be better to check the event key first and act on it only on the events of interest, no?

operramon: let me rethink it

operramon: Fixed

mpellerin42 (approved):

PR #1255, opened 2024-01-11
[sc-8274] consistent language model on KB creation

mpellerin42:
- Create a standalone component for language field in KB creation
- Move function allowing to get KB configuration from backend configuration and user’s choices to the core lib.
- Update KB création dialog to use the new language-field component
- Fix KB creation from onboarding by passing account id instead of account slug (have been broken during SDK clean up)

### DIFF
--- a/libs/common/src/lib/kb-add/kb-add.component.ts
+++ b/libs/common/src/lib/kb-add/kb-add.component.ts
@@ -156,7 +156,6 @@ export class KbAddComponent implements OnInit {
   }
 
   updateLanguages(data: { multilingualSelected: boolean; languages: string[] }) {
-    console.log(`updateLanguages`, data);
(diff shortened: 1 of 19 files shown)


libs/common/src/lib/kb-add/kb-add.component.ts
```diff
@@ -140,8 +155,10 @@ export class KbAddComponent implements OnInit {
     this.cdr?.markForCheck();
   }
 
-  hasTranslation(key: string) {
-    const translation = this.translate.instant(key);
-    return translation !== key && translation !== '';
+  updateLanguages(data: { multilingualSelected: boolean; languages: string[] }) {
+    console.log(`updateLanguages`, data);
```
ebrehault, line 159: to be cleaned

ebrehault (approved):

PR #1266, opened 2024-01-19
Ericbrehault/sc 8410/support sitemap in sync agent

ebrehault:


operramon (approved):

libs/sync/src/lib/upload/select-files/select-files.component.ts
```diff
@@ -138,8 +138,8 @@ export class SelectFilesComponent implements AfterViewInit {
         // Add empty item just to trigger sitemap synchronization
         this.selection.setSelection({
           uuid: '',
-          title: '',
-          originalId: '',
+          title: 'Sitemap',
+          originalId: 'SITEMAP',
```
operramon, line 142: should the desktop app version number be increased ?

ebrehault: no, that's for the synchronizer

PR #1276, opened 2024-01-26
[sc-8642] Maintenance page

mpellerin42:


operramon (approved):

ebrehault (approved):
We will need a corresponding route in the nginx config but we can still merge this PR and fix that later (after discussing with Platform team what they prefer)

PR #1282, opened 2024-01-30
[sc-8640] Support RAG strategies options in the widget

mpellerin42:
- Add support for RAG strategies in the SDK
- Add `rag_strategies` and `rag_field_ids` options to the widget
- Add support for RAG strategies options in the widget editor page
- Fix bug in widget generator advanced tab: `edited` marker is now displayed only when the preset is really modified


libs/search-widget/src/widgets/search-widget/SearchBar.svelte
```diff
@@ -47,6 +52,8 @@
   export let cssPath = '';
   export let prompt = '';
   export let no_tracking = false;
+  export let rag_strategies = '';
```
operramon, line 55: should these options also be added to ChatWidget.svelte ?

operramon (approved):

libs/search-widget/src/widgets/search-widget/SearchBar.svelte
mpellerin42: good question, I don't know but I'll ask @ebrehault

PR #1286, opened 2024-01-30
render wild labels when editing resource

ebrehault:


operramon (approved):

libs/common/src/lib/resources/edit-resource/classification/resource-classification.component.ts
operramon, line 1: If a resource have two wild labels, only one is checked as selected, is this expected ?

ebrehault: it is not, let me check

PR #1299, opened 2024-02-05
[sc-8738] Support backend feature flags

mpellerin42:
- Add support for backend feature flag
- Use NucliaDB flag for new processing status endpoint

### DIFF
--- a/libs/core/src/lib/analytics/feature-flag.service.ts
+++ b/libs/core/src/lib/analytics/feature-flag.service.ts
@@ -20,6 +20,7 @@ interface FeaturesData {
 
 const CUSTOM_FEATURE_FLAGS = 'NUCLIA_CUSTOM_FEATURE_FLAGS';
 const FEATURE_PREFIX = 'application_';
+const BACKEND_PREFIXES = ['nucliadb_', 'nua_'];
 
 const stageFeatures: Features = {};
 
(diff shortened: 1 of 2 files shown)


libs/core/src/lib/analytics/feature-flag.service.ts
```diff
@@ -47,8 +48,23 @@ export class FeatureFlagService {
...
+            ...acc,
+            [key]: feature?.rollout === 100 || (feature?.variants?.account_id_md5 || []).includes(md5 || ''),
+          };
+        }, {}),
+    ),
+  );
   private stageFeatures = new BehaviorSubject<Features>({ ...stageFeatures, ...this.getCustomFeatures() });
-  private features: Observable<Features> = this.remoteFeatures.pipe(
+  private applicationFeatures: Observable<Features> = this.applicationRemoteFeatures.pipe(
+    map((features) => ({ ...features, ...this.getCustomFeatures() })),
```
operramon, line 65: I guess this line is not needed, because at the moment backend flags are not customizable with the konami code

operramon (approved):

libs/core/src/lib/analytics/feature-flag.service.ts
mpellerin42: ok, I wasn't sure what those customFeatures were about. Let me clean that up

PR #1304, opened 2024-02-06
[sc-8765] Update Notification panel

mpellerin42:
- Use the new `listenToResourceOperationNotifications` method to get better notifications
- Stop listening to current KB notifications when switching to another KB, and remove the previous notifications from the panel
- Fix the error management in rest’s `getStreamMessages` method: firefox is raising `NS_BINDING_ABORTED` TypeError when reloading the page while the connection is opened, so our system allowing to reconnect once in case of network error was reconnecting while it shouldn't have on this specific case.


ebrehault (approved):

libs/common/src/lib/notifications/notification.service.ts
```diff
@@ -22,8 +24,15 @@ export class NotificationService {
   ) {
     combineLatest([this.sdk.currentAccount, this.sdk.currentKb])
       .pipe(
+        tap(([, kb]) => {
```
operramon, line 27: If I logout and login again, it seems the notifications connection is lost. This also happens when navigating to the user profile. Are these edge cases worth fixing ?

operramon (approved):

libs/common/src/lib/notifications/notification.service.ts
mpellerin42: yes they do, thank you for raising those issues, I'll fix them before merging :)

operramon (approved):

PR #1312, opened 2024-02-09
[sc-8785] and [sc-8784]: Fix folder and sitemap connectors

mpellerin42:
- Don't show folders tab if canSelectFiles is false
- Fix folder and sitemap connectors


ebrehault (approved):
We can merge it if you want to move on, but I think it would be cleaner to have a `allowToSelectFiles` attributes on the connectors, that we do not have to hard a test about is it `folder` or `sitemap` and we do not have to test if your new `getItems` method exists or not, we just check is `allowToSelectFiles` is true.
It would be easier to understand why in some cases we call getItems.

It can be done later if you prefer.

mpellerin42:
> We can merge it if you want to move on, but I think it would be cleaner to have a `allowToSelectFiles` attributes on the connectors, that we do not have to hard a test about is it `folder` or `sitemap` and we do not have to test if your new `getItems` method exists or not, we just check is `allowToSelectFiles` is true. It would be easier to understand why in some cases we call getItems.
> 
> It can be done later if you prefer.

well, it's already what we do: in `add-sync.component` we check `canSelectFiles` which is an equivalent to `allowToSelectFiles` don't you think? We also check `getItems` type just because it's optional in the interface (like `handleParameters` for instance).

Also we're not checking for `sitemap` or `folder` anywhere any more, so I don't really understand this part of your comment.
In any case, there’s no rush on merging this one as I found out I didn't fully fix sitemap connector, so we can discuss this further if you want.

ebrehault (approved):

PR #1313, opened 2024-02-09
[sc-8798] Check sync server status

mpellerin42:
- Check sync server status every 5s when it is down, and every 5min when it is up
- Push a notification when the server is down
- Move notifications folder to `core` lib to prevent circle dependencies


libs/core/src/lib/notifications/notification.service.ts
```diff
@@ -2,7 +2,7 @@ import { Injectable } from '@angular/core';
 import { BehaviorSubject, combineLatest, filter, map, Observable, switchMap, tap } from 'rxjs';
 import { NotificationData, NotificationUI } from './notification.model';
 import { SDKService } from '@flaps/core';
-import { NavigationService } from '../services';
+import { NavigationService } from '../../../../common/src/lib/services';
```
ebrehault, line 5: bad import

ebrehault (approved):

PR #1314, opened 2024-02-09
[sc-8799] Format summary displayed in resource preview

operramon:


mpellerin42 (approved):
this solution is working for sure, but I'm wondering if it's really the most performant. I feel like we have this kind of line break translation quite often, maybe it would worth considering a simple pipe replacing all `\n` by `<br>` and using an `innerHtml` to display the field?
something like

```html
<p [innerHtml]="(resource.summary | lineBreakTranslator) || '–'"></p>
```

ebrehault (approved):

PR #1319, opened 2024-02-15
[sc-8851] Error on summarize and predictSummarize

operramon:


libs/sdk-core/src/lib/db/db.ts
```diff
@@ -552,8 +568,13 @@ export class Db implements IDb {
...
     }
     const modelParam = model ? `?model=${encodeURIComponent(model)}` : '';
-    return this.nuclia.rest
-      .post<{ summary: string }>(
+    const retryDelays = [0, 1000, 2000, 5000, 10000];
+    const retryConfig: RetryConfig = {
+      count: retryDelays.length,
+      delay: (error, retryCount) => timer(retryDelays[retryCount - 1]),
+    };
+    return defer(() =>
```
mpellerin42, line 576: why do you need a `defer` here?

mpellerin42 (approved):

libs/sdk-core/src/lib/db/db.ts
operramon: I had to add it because otherwise the `retry` operator didn't work as expected.
I got the idea from here: https://stackoverflow.com/questions/33072512/rx-frompromise-and-retry
maybe in the future we can find out a better solution.

PR #1335, opened 2024-02-20
[sc-8950] Prompt lab improvements

mpellerin42:
- allow only one prompt and keep it displayed in the textarea
- Use expander at question level instead of model level
- Use card expander within a container with a border to nicely separate the questions


libs/common/src/lib/prompt-lab/prompt-lab.component.ts
```diff
@@ -61,15 +63,126 @@ export class PromptLabComponent implements OnInit {
...
-  prompts: string[] = [];
 
   configBackup?: { [id: string]: any };
   learningModels?: LearningConfiguration;
   loadingModal?: ModalRef;
   progress$ = new BehaviorSubject<number | null>(null);
-  results: { query: string; data: { prompt?: string; results: { model: string; answer: string }[] }[] }[] = [];
-  hasResults = new BehaviorSubject(false);
-  modelCollapsed: { [id: string]: boolean } = {};
+  results: { query: string; data: { prompt?: string; results: { model: string; answer: string }[] }[] }[] = [
```
ebrehault, line 71: why do we need hard-coded results?

ebrehault (approved):

libs/common/src/lib/prompt-lab/prompt-lab.component.ts
mpellerin42: we don't, I just forget to remove them 🙄

PR #1336, opened 2024-02-20
[sc-8778] Improve the sync server settings

operramon:


libs/common/src/lib/upload/upload-data/upload-layout.component.html
```diff
@@ -1,8 +1,14 @@
...
-@if (!standalone && hasSyncServer) {
+@if (!standalone && (hasSyncServer | async)) {
   <div class="sync-status">
     {{ ((isServerDown | async) ? 'upload.server.server_down' : 'upload.server.server_up') | translate }}
+    <pa-button
+      (click)="setupSync()"
+      icon="gear"
+      size="small"
+      aspect="basic"
+      paTooltip="stash.upload.sync.setup-button"></pa-button>
```
mpellerin42, line 12: for accessibility purpose, you should always add the label as well on any button (even if it's not visible), as it's automatically becoming an `aria-label` on icon buttons:
```suggestion
      paTooltip="stash.upload.sync.setup-button">{{ 'stash.upload.sync.setup-button' | translate }}</pa-button>
```

mpellerin42 (approved):

PR #1357, opened 2024-02-27
display logs for curent sync

ebrehault:


operramon (approved):

libs/sync/src/lib/sync/new-sync.service.ts
operramon, line 1: It seems there's a `this.sdk.currentKb` without `take(1)` in the `setSourceAndDestination` method

PR #1361, opened 2024-03-01
Ericbrehault/sc 8796/fix sync agent support in nucliadb admin

ebrehault:


apps/dashboard/src/app/app.module.ts
```diff
@@ -37,10 +37,10 @@ import {
 import { KnowledgeBoxModule } from './knowledge-box';
 import { InviteModule } from './invite/invite.module';
 import { MultiTranslateHttpLoader } from 'ngx-translate-multi-http-loader';
-import { SyncService } from '@nuclia/sync';
-import { NewSyncService } from '../../../../libs/sync/src/lib/sync/new-sync.service';
 import { TitleStrategy } from '@angular/router';
 import { AppTitleStrategy } from './app-title.strategy';
+import { SyncService } from '@nuclia/sync';
+import { NewSyncService } from 'libs/sync/src/lib/sync/new-sync.service';
```
mpellerin42, line 43: Will `NewSyncService` fully replace `SyncService` at some point? is this why we keep it out of `@nuclia/sync`?

libs/common/src/lib/navbar/navbar.component.ts
```diff
@@ -23,7 +23,7 @@ export class NavbarComponent implements OnInit, OnDestroy {
   );
   inSettings: Observable<boolean> = combineLatest([this.sdk.currentAccount, this.sdk.currentKb]).pipe(
     map(([account, kb]) => {
-      return this.navigationService.getKbUrl(account.slug, kb.slug || '');
+      return this.navigationService.getKbUrl(account.slug, this.standalone ? kb.id : kb.slug || kb.id);
```
mpellerin42, line 26: you have `this.standalone ? kb.id : kb.slug || kb.id` twice in the code, maybe it would worth extracting it in its own property?

libs/sync/src/lib/upload/sync.component.ts
```diff
@@ -27,10 +28,16 @@ export class SyncComponent implements OnInit, OnDestroy {
...
+    private sdk: SDKService,
   ) {}
 
   ngOnInit() {
-    this.syncService.setCurrentSourceId(location.pathname.split('/upload/sync/')[1]?.split('/')[0]);
+    this.syncService.setCurrentSourceId(
+      (this.sdk.nuclia.options.standalone ? location.hash : location.pathname)
+        .split('/upload/sync/')[1]
+        ?.split('/')[0]
+        .split('?')[0] || '',
```
mpellerin42, line 39: this code is hard to read and might be are to debug in case of problem later. I don't really understand why you use global location instead of angular’s ActivatedRoute, it seems like it would be safer and clearer to get the source Id from there.

apps/dashboard/src/app/app.module.ts
ebrehault: yes, as soon as desktop is gone

libs/sync/src/lib/upload/sync.component.ts
```diff
@@ -28,16 +27,13 @@ export class SyncComponent implements OnInit, OnDestroy {
...
   ngOnInit() {
-    this.syncService.setCurrentSourceId(
-      (this.sdk.nuclia.options.standalone ? location.hash : location.pathname)
-        .split('/upload/sync/')[1]
-        ?.split('/')[0]
-        .split('?')[0] || '',
-    );
+    this.route.paramMap.subscribe((params) => {
+      this.syncService.setCurrentSourceId(params.get('id') || '');
+    });
```
mpellerin42, line 36: this feels so nice and clean now 😍

mpellerin42 (approved):

libs/sync/src/lib/sync/new-sync.service.ts
```diff
@@ -211,53 +212,44 @@ export class NewSyncService {
...
+                zone: this.sdk.nuclia.options.zone,
+                backend: this.sdk.nuclia.options.backend,
+                knowledgeBox: data.kbid,
+                apiKey: data.token,
+              } as NucliaOptions,
+            })),
+          );
+        }
+      }),
+      switchMap((source) => this.setSourceData(sourceId, source)),
```
operramon, line 244: when running `setSourceData`, the sync agent returns `Invalid format for kb: Error: zone: Required, apiKey: Required`, Is this expected ?

operramon (approved):

libs/sync/src/lib/sync/new-sync.service.ts
ebrehault: I forgot to push my sync-agent PR :)

PR #1364, opened 2024-03-01
backward compat for new and old subscriptions

ebrehault:


libs/core/src/lib/api/billing.service.ts
```diff
@@ -90,6 +92,19 @@ export class BillingService {
     );
   }
 
+  getStripeSubscription(): Observable<StripeAccountSubscription | null> {
+    return this.getSubscription().pipe(
+      map((subscription) => {
+        if (!(subscription as unknown as any).provider) {
+          return null;
```
mpellerin42, line 99: you should not return null in this case, is there is no provider it's always a `StripeSubscription`
```suggestion
          return subscription;
```

mpellerin42 (changes requested):

libs/core/src/lib/api/billing.service.ts
```diff
@@ -90,6 +92,19 @@ export class BillingService {
...
+        } else {
+          return (subscription as AccountSubscription).subscription as StripeAccountSubscription;
+        }
+      }),
+      catchError(() => of(null)),
+    );
+  }
```
mpellerin42, line 106: Actually with the typing from `getSubscription`, there's a lot of cast that you don't need here:

```suggestion
  getStripeSubscription(): Observable<StripeAccountSubscription | null> {
    return this.getSubscription().pipe(
      map((subscription) => {
        if (!subscription.provider) {
          return subscription;
        } else {
          return subscription.subscription;
        }
      }),
      catchError(() => of(null)),
    );
  }
```

mpellerin42 (approved):

PR #1386, opened 2024-03-07
[sc-9217] Fix aws account detection

operramon:


libs/core/src/lib/api/billing.service.ts
```diff
@@ -38,13 +38,7 @@ export class BillingService {
...
   isSubscribedToAws = this.sdk.nuclia.options.standalone
     ? of(false)
-    : this.type.pipe(
-        switchMap((type) =>
-          type === 'v3enterprise'
-            ? this.getSubscription().pipe(map((subs) => subs?.provider === 'AWS_MARKETPLACE'))
-            : of(false),
-        ),
-      );
+    : this.type.pipe(switchMap(() => this.getSubscription().pipe(map((subs) => subs?.provider === 'AWS_MARKETPLACE'))));
```
mpellerin42, line 41: you're not using the `type` anymore, do you? so is it really needed to first call it here? couldn't we direclty do
```suggestion
    : this.getSubscription().pipe(map((subs) => subs?.provider === 'AWS_MARKETPLACE'));
```
?

mpellerin42 (approved):

ebrehault (approved):

PR #1391, opened 2024-03-08
[sc-8823] Extend ability to delete all resources

operramon:


libs/common/src/lib/resources/resource-list/pending-resources-table/pending-resources-table.component.ts
```diff
@@ -11,6 +13,8 @@ import { RESOURCE_STATUS } from '@nuclia/core';
 })
 export class PendingResourcesTableComponent extends ResourcesTableDirective {
   override status: RESOURCE_STATUS = RESOURCE_STATUS.PENDING;
+  private uploadService = inject(UploadService);
+  totalCount = this.uploadService.statusCount.pipe(map((statusCount) => statusCount.pending));
```
mpellerin42, line 17: Not sure but I feel like this statusCount is already known in the ResourceList parent, if so maybe we could have this `totalCount` defined as an input.

mpellerin42 (approved):

libs/common/src/lib/resources/resource-list/pending-resources-table/pending-resources-table.component.ts
operramon: let me check...

PR #1394, opened 2024-03-08
[sc-9066] Fix Konami features modal

mpellerin42:


libs/common/src/lib/features/features-modal.component.html
```diff
@@ -1,19 +1,22 @@
-<pa-modal-advanced>
+<pa-modal-advanced [fitContentHeight]="!!features">
   <pa-modal-title>{{ 'features.customize_features' | translate }}</pa-modal-title>
   <pa-modal-content>
-    <ng-container *ngIf="features">
-      <div *ngFor="let feature of features | keyvalue">
+    @if (features; as features) {
+      @for (feature of features | keyvalue; track feature) {
         <pa-checkbox [(value)]="features[feature.key]">{{ feature.key }}</pa-checkbox>
```
operramon, line 6: When enabling a feature, and then opening the modal again, it seems this checkbox is unchecked. Is this expected ?

operramon (approved):

libs/common/src/lib/features/features-modal.component.html
mpellerin42: did you hit save? when I tried, hitting save was reloading the page and the checkbox was enabled. then clicking on reset put it back as it was (all checkbox disabled)

mpellerin42: fixed

PR #1395, opened 2024-03-08
Fix routing after creating KB from KB selection page

mpellerin42:


libs/common/src/lib/select-account-kb/select-kb/select-kb.component.ts
```diff
@@ -72,7 +72,7 @@ export class SelectKbComponent implements OnDestroy {
       )
       .subscribe((result) => {
         this.sdk.refreshKbList();
```
operramon, line 74: I could reproduce the zone issue, I guess in some cases the `nuclia.options.zone` is wrong here.

operramon (approved):

PR #1397, opened 2024-03-11
[sc-9236] Restrict to ChatGPT 3.5 for Starter accounts

mpellerin42:


libs/sdk-core/src/lib/db/kb/kb.ts
```diff
@@ -58,6 +57,22 @@ const TEMP_TOKEN_DURATION = 5 * 60 * 1000; // 5 min
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
 export interface KnowledgeBox extends IKnowledgeBox {}
 
+function getOnlyChatGpt3Model(allSchema: LearningConfigurations): LearningConfigurations {
```
ebrehault, line 60: It's not great to have that in the sdk (as it reflects a concern of us as a company, while the sdk is meant to be a neutral tool to use the Nuclia API), but I guess that's the easiest way. So let's go with that.

ebrehault (approved):

PR #1405, opened 2024-03-12
[sc-9016] Use any operator in the resources list filtering

operramon:


libs/sdk-core/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Allow to modify the HTTP headers in the Nuclia REST API calls.
 - Add support for `css_selector` in Link fields
+- Support POST requests for `/catalog` endpoint
```
mpellerin42, line 7: 1.14.0 version has been released today in a previous PR but I forgot to add the date in the CHANGELOG. Can you add it and create a new `1.14.1` entry please?

libs/sdk-core/src/lib/db/search/search.ts
```diff
@@ -103,13 +103,14 @@ export const search = (
   return manageSearchRequest(nuclia, kbid, searchMethod);
 };
 
-export const catalog = (nuclia: INuclia, kbid: string, query: string, options?: SearchOptions) => {
+export const catalog = (nuclia: INuclia, kbid: string, query: string, options?: SearchOptions, useGet?: boolean) => {
```
mpellerin42, line 106: I don't understand: why did you create this `useGet` option while you're not using it anywhere in your PR 🤔

ebrehault: @mpellerin42 if `/catalog` supports POST now, I think it is nice to offer this through the SDK (even if we do not use it in the dashboard, it could be used by external users)

ebrehault (approved):

libs/sdk-core/CHANGELOG.md
operramon: Fixed

mpellerin42 (approved):

PR #1415, opened 2024-03-14
Ericbrehault/sc 8659/clean up desktop

ebrehault:


libs/common/src/lib/utils/desktop.ts
```diff
@@ -38,24 +38,6 @@ export function getDesktopAppUrl(platform: 'mac' | 'win' | 'linux'): Observable<
...
-  timer(400).subscribe(() => {
-    if (!appInstalled) {
-      downloadDesktop();
-    }
-    window.removeEventListener('blur', onBlur);
-  });
-  window.location.href = 'nuclia-desktop://';
-}
-
 function downloadDesktop() {
```
mpellerin42, line 41: This file is named desktop.ts, do we still need it? If so, we should rename it. Do we still need this downloadDesktop method?

libs/sync/src/lib/sync/sync.service.ts
```diff
@@ -343,144 +321,106 @@ export class SyncService {
...
+              ({
+                date: log.createdAt,
+                from: log.payload?.['title'],
+                to: log.payload?.['kb']?.['knowledgeBox'],
+                total: 1,
+                progress: 1,
+                started: true,
+                completed: true,
+                level: log.level,
+                errors: log.message, // we use the message field to store the error, but when dropping desktop, we will not use this mapping anymore
```
mpellerin42, line 393: This comment seems to indicate this mapping isn't useful for syncs

operramon (approved):

mpellerin42 (approved):

PR #1428, opened 2024-03-19
[sc-9361] Add management of allowed CORS domains in KB settings page

mpellerin42:
The frontend is fully ready, but resetting the field doesn't work yet (Core services team is aware and working on a fix)

### DIFF
--- a/libs/sdk-core/src/lib/db/kb/kb.models.ts
+++ b/libs/sdk-core/src/lib/db/kb/kb.models.ts
@@ -35,6 +35,7 @@ export interface IKnowledgeBoxCreation {
   description?: string;
   zone: string;
   uuid?: string;
+  allowed_origins?: string[] | null;
 }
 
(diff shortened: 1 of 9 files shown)


libs/sdk-core/src/lib/db/kb/kb.models.ts
```diff
@@ -35,6 +35,7 @@ export interface IKnowledgeBoxCreation {
   description?: string;
   zone: string;
   uuid?: string;
+  allowed_origins?: string[] | null;
```
ebrehault, line 38: why `null` here?
the question mark make it already possibly `undefined`

ebrehault (approved):

libs/sdk-core/src/lib/db/kb/kb.models.ts
mpellerin42: because we need to pass `null` to reset the field.

PR #1435, opened 2024-03-21
Revert search page/widget editor merging

ebrehault:


apps/sistema-demo/src/e2e/prod/search-europe.html
operramon, line 1: would removing this file break e2e tests ?

apps/sistema-demo/src/e2e/search.html
operramon, line 1: would removing this file break e2e tests ?

operramon (approved):

apps/sistema-demo/src/e2e/search.html
ebrehault: oh I just noticed the search-usa one and forgot about the rest, let me fix it

PR #1440, opened 2024-03-25
Display prompt from settings and show if applied

ebrehault:


mpellerin42 (approved):

libs/common/src/lib/widgets/widget-generator.component.ts
```diff
@@ -279,6 +281,7 @@ export class WidgetGeneratorComponent implements OnInit, OnDestroy {
       .subscribe(([schema, config]) => {
         this.generativeModels = schema['generative_model']?.options || [];
         this.defaultModelFromSettings = config['generative_model'] || '';
+        this.defaultPromptFromSettings = config['user_prompts']?.[this.defaultModelFromSettings]?.['prompt'] || '';
```
operramon, line 284: Maybe there's a problem in this here.
It seems like the models keys are not very consistent in the configuration.
For instance: `config['generative_model']` is `claude-3`
but the user prompt is stored in `config['user_prompts']['claude3']`

operramon (approved):

libs/common/src/lib/widgets/widget-generator.component.ts
ebrehault: good catch!
fixed

PR #1461, opened 2024-04-04
[sc-9472] Enhance citations experience

operramon:


libs/common/src/lib/widgets/widget-generator.component.ts
```diff
@@ -840,4 +840,10 @@ ${baseSnippet.replace('zone=', copiablePrompt + 'zone=')}`;
         this.defaultModelFromSettings === this.advancedForm.controls.generativeModel.value) &&
       this.userPrompt === '';
   }
+
+  onToggleCitations(useCitations: boolean) {
+    if (useCitations) {
+      this.hideResultsControl.patchValue(true);
+    }
+  }
```
mpellerin42, line 848: I feel like this is not enough: shouldn't we also put back `hideResults` to false when we disable citations?

operramon: I was not sure,
when `citations` is switched off, we don't know if `hide results` was enabled because of the citations, or because the user manually set it. Should we disable`hide results` automatically in any case ?

ebrehault: I guess that's good enough. Anyway, the new citations are coming (I think), so when it will be more obvious they are citations, we will just let results on.

ebrehault (approved):

PR #1474, opened 2024-04-08
Sync security groups

ebrehault:


libs/sync/src/lib/upload/tabs/edit-settings.component.ts
```diff
@@ -54,6 +58,7 @@ export class EditSyncSettingsComponent implements OnInit {
                 ...sync.connector,
                 parameters,
               },
+              syncSecurityGroups: this.form?.value['syncSecurityGroups'],
```
operramon, line 61: is there a PR in sync agent repo ?

operramon (approved):

libs/sync/src/lib/upload/tabs/edit-settings.component.ts
ebrehault: yes, but I first need to merge this one, so I get the latest sdk released on NPM, so I can use it in the sync agent

PR #1486, opened 2024-04-12
Ericbrehault/sc 9596/dump current query

ebrehault:


libs/sdk-core/src/lib/events.ts
```diff
@@ -1,18 +1,29 @@
...
-  on<T>(eventName: string): Observable<T> {
+  on<T = any>(eventName: string): Observable<T> {
     return this.messages.asObservable().pipe(
       filter((message) => message.name === eventName),
       map((message) => message.data as T),
     );
   }
+
+  log(eventName: string, data: any): void {
+    this.logs.next({ ...this.logs.getValue(), [eventName]: data });
```
mpellerin42, line 23: when does the logs reset? won't they grow forever with this code?

libs/search-widget/public/i18n/en.json
```diff
@@ -1,5 +1,6 @@
 {
 	"answer.chat-action": "Ask more",
+	"answer.download-log": "Download log",
```
mpellerin42, line 3: shouldn't this be plural (logs)? (if so you need to update all the translations)

libs/search-widget/src/core/utils.ts
```diff
@@ -385,3 +385,13 @@ export function hasNotEnoughData(text: string): boolean {
   text = text.replace(/<br>/g, '').replace(/\n/g, '');
   return text === 'Not enough data to answer this.';
 }
+
+export function downloadAsJSON(data: any) {
+  const element = document.createElement('a');
+  element.setAttribute('href', 'data:application/json;charset=utf-8,' + encodeURIComponent(JSON.stringify(data)));
+  element.setAttribute('download', `dump-${new Date().toISOString()}.json`);
+  element.style.display = 'none';
+  document.body.appendChild(element);
```
mpellerin42, line 394: I don't think you need to add the element in the body to be able to click on it.

libs/sdk-core/src/lib/events.ts
ebrehault: it's not a list, just a dict, so for any given key, we override the previous value on each change

libs/search-widget/public/i18n/en.json
ebrehault: it is just the last log entry

libs/sdk-core/src/lib/events.ts
mpellerin42: oh right

mpellerin42 (approved):

operramon (approved):

PR #1493, opened 2024-04-16
Ericbrehault/sc 9683/set targeted resource in widget

ebrehault:


libs/search-widget/src/components/answer/Answer.svelte
```diff
@@ -73,15 +73,15 @@
           aspect="basic"
           size="small"
           on:click={() => dispatch('openChat')}>
-          <span class="go-to-chat title-s">{$_('answer.chat-action')}</span>
```
operramon, line 76: I remember `go-to-chat` is used in e2e, could it break some test ?

operramon (approved):

PR #1495, opened 2024-04-17
processing options when uploading files

ebrehault:


libs/common/src/lib/upload/upload-files/upload-files.component.ts
```diff
@@ -105,12 +110,16 @@ export class UploadFilesComponent {
...
+      let labelledFiles = this.setLabels(files);
       if (this.langCode) {
         labelledFiles.forEach((file) => {
           file.lang = this.langCode.getRawValue();
         });
       }
+      if (this.processing !== 'none') {
+        labelledFiles.forEach((file) => (file.contentType = `${file.type}+${this.processing}`));
+      }
+      console.log(labelledFiles);
```
operramon, line 122: console.log ?

operramon (approved):

libs/common/src/lib/upload/upload-files/upload-files.component.ts
ebrehault: oops

PR #1507, opened 2024-04-22
[sc-9791] Hide the API info boxes when not trial account

operramon:


ebrehault (approved):

mpellerin42 (approved):
This is OK for now, but I think we duplicate too much the code allowing to check is we ar in trial or not, we have it spread everywhere in the apps. We should probably centralise this check in a service. I'll create a new ticket for that.

PR #1514, opened 2024-04-29
Support max_tokens

ebrehault:


libs/sdk-core/src/lib/db/kb/kb.ts
```diff
@@ -296,6 +296,7 @@ export class KnowledgeBox implements IKnowledgeBox {
     options?: ChatOptions,
     callback?: (answer: Chat.Answer | IErrorResponse) => void,
   ): Observable<Chat.Answer | IErrorResponse> | Observable<null> {
+    console.log('chat', query, context, features, options);
```
mpellerin42, line 299: shouldn't you clean up this log?

mpellerin42 (approved):

libs/sdk-core/src/lib/db/kb/kb.ts
ebrehault: oops

PR #1526, opened 2024-05-03
[sc-10040] Highlight citations on hover

operramon:


libs/search-widget/src/components/answer/Answer.svelte
```diff
@@ -34,6 +35,11 @@
   $: sources =
     answer.citations && answer.sources?.resources ? getSourcesResults(answer.sources?.resources, answer.citations) : [];
 
+  $: sources && document.querySelectorAll('.sw-answer .ref').forEach((ref) => {
+    ref.addEventListener('mouseenter', onMouseEnter);
+    ref.addEventListener('mouseleave', onMouseLeave);
```
mpellerin42, line 40: you add event listener anytime the sources are changing, can that happens several times? I see you never remove the listeners, did you check each of them were not called several times?

operramon: I just check it, and it seems the event handler is called one time.
At first I used `removeEventListener`, but then I realized that calling `addEventListener` with the exact same parameters doesn't add an extra listener. I can modify it if it's not clear.

mpellerin42: if you checked and there is no multiple listeners added on the same element, then no problem :)

mpellerin42 (approved):

PR #1538, opened 2024-05-13
support markdown in results

ebrehault:


libs/search-widget/src/common/paragraph-result/ParagraphResult.svelte
```diff
@@ -82,9 +85,11 @@
     </div>
     <div
       class="paragraph-text"
-      class:ellipsis={ellipsis && !expanded}
+      class:ellipsis={hasEllipsis && !expanded}
       bind:this={paragraphElement}>
-      {@html paragraph.text}
+      <MarkdownRendering
```
operramon, line 90: is it ok to render all paragraphs as markdown ? or should only the fields with "KEEP_MARKDOWN" format be rendered ?

operramon (approved):

libs/search-widget/src/common/paragraph-result/ParagraphResult.scss
```diff
@@ -38,6 +40,19 @@
...
+        height: var(--rhythm-3);
+        :global(.markdown > *) {
+          // artificial ellipsis if the paragraph has several lines
+          &::after {
+            content: '…';
+          }
+          // regular ellipsis for single line that is too long
+          overflow: hidden;
+          text-overflow: ellipsis;
+          white-space: nowrap;
```
mpellerin42, line 53: @ebrehault don't we risk to have 2 ellipsis in certain edge case?

PR #1544, opened 2024-05-16
put rephrase in base search option

ebrehault:


mpellerin42 (approved):

mpellerin42 (commented):
Should we add a line in the CHANGELOG as it's in the SDK?

PR #1554, opened 2024-05-28
Ericbrehault/sc 10019/sync agent bugs

ebrehault:


libs/sync/src/lib/add-sync-page/add-sync-page.component.ts
```diff
@@ -211,6 +211,7 @@ export class AddSyncPageComponent implements OnInit {
   }
 
   updateSelection(selection: SyncItem[]) {
+    console.log('selection', selection);
```
operramon, line 214: is it needed ?

operramon (approved):

PR #1564, opened 2024-06-03
[sc-10062] Add dialog to create missing label sets

operramon:


libs/common/src/assets/i18n/fr.json
```diff
@@ -586,6 +586,11 @@
 	"label-set.form.name": "Nom du jeu d'étiquettes",
 	"label-set.form.name-invalid": "Nom du jeu d'étiquettes non valide",
 	"label-set.form.name-required": "Le nom du jeu d'étiquettes est obligatoire",
+	"label-set.missing.check-missing": "Vérifier les étiquettes manquantes",
+	"label-set.missing.declare": "Déclarer",
+	"label-set.missing.help": "Si vous avez utilisé CLI/API pour créer des étiquettes, vous devez les déclarer.",
+	"label-set.missing.labels-detected": "Nous avons détecté certains jeux d'étiquettes générés via la CLI/API. Consultez la liste et déclarez les jeux d'étiquettes.",
+	"label-set.missing.no-labels-detected": "Aucun jeu de balises non déclaré n’a été détecté.",
```
mpellerin42, line 593: "jeu de balises" > `jeu d'étiquettes`

libs/core/src/lib/label/label-sets/label-sets.component.ts
```diff
@@ -28,11 +31,23 @@ export class LabelSetsComponent {
   selectionLabelSets: Observable<LabelSetDisplay[]> = this._labelSets.pipe(
     map((labelSets) => labelSets.filter((labelSet) => labelSet.kind.includes(LabelSetKind.SELECTIONS))),
   );
+  showingLabelSetForm = this.router.events.pipe(
+    filter((event) => event instanceof NavigationEnd),
+    startWith(() => true),
+    map(() => this.route.children.length > 0),
+  );
```
mpellerin42, line 38: it's not obvious and a bit hard to understand the causality between the route’s children and the fact we show label set form. Is there any simpler approach?

mpellerin42 (approved):

libs/common/src/assets/i18n/fr.json
operramon: Fixed

libs/core/src/lib/label/label-sets/label-sets.component.ts
operramon: oh, yes, it's not very obvious.
The only idea that comes to mind is to use `location.pathname.endsWith('/label-sets') === false`.
would it be clearer ?

mpellerin42: could you achieve the same using `activatedRoute` instead of location? It's always better to stick with Angular routing component than browser location.

operramon: ok, now it's fixed

PR #1568, opened 2024-06-07
[sc-10181] implement new search and widgets sections

mpellerin42:
New search page in the Dashboard (and NucliaDB admin).

The new widget page will be done as a second step, so currently the "Create Widget" button is just redirecting to the actual Widget generator page.


ebrehault (approved):

libs/core/src/lib/unauthorized-feature/unauthorized-feature.directive.ts
```diff
@@ -44,6 +44,7 @@ export class UnauthorizedFeatureDirective {
   }
 
   @HostListener('click', ['$event']) onClick(event: MouseEvent) {
+    console.log(`click`);
```
operramon, line 47: is it needed ?

operramon (approved):

libs/core/src/lib/unauthorized-feature/unauthorized-feature.directive.ts
mpellerin42: definitely not, good catch

PR #1570, opened 2024-06-07
[sc-10135] Re-build widget for nuclia.com to use ask endpoint

operramon:


mpellerin42 (commented):
Why did you copy `common-style.scss` and `_global.scss`? I think those two should be reused by just referencing them where needed, no?

mpellerin42 (approved):
I feel like it's ok, just wondering if it's needed to duplicate `LoadingDots` as well: duplicating components that have been modified for the specificity of GolbalWidget makes sense, but any components which is just the same as the common one should not be duplicated.

operramon:
> Why did you copy `common-style.scss` and `_global.scss`? I think those two should be reused by just referencing them where needed, no?

ah, I thought the global styles were different, but I see that they are the same.
Now it's fixed.

operramon:
> I feel like it's ok, just wondering if it's needed to duplicate `LoadingDots` as well: duplicating components that have been modified for the specificity of GolbalWidget makes sense, but any components which is just the same as the common one should not be duplicated.

Fixed

ebrehault (approved):

PR #1571, opened 2024-06-10
Ericbrehault/sc 10359/unit tests on sync agent flow

ebrehault:


libs/sync/src/lib/sync-details-page/sync-details-page.component.html
```diff
@@ -18,13 +18,15 @@ <h1 class="page-title">
               <pa-toggle
                 withBackground
                 [value]="!sync.disabled"
+                qa="active-toggle"
```
mpellerin42, line 21: following what we learned while implementing the e2e tests, using custom attributes like `qa` is not a good practice. Why not using the same as what we use for cypress: `data-cy` ? this way we could reuse them in e2e tests at some point if needed.

mpellerin42 (approved):

operramon (approved):

PR #1583, opened 2024-06-13
support the new schemas endpoint

ebrehault:


libs/common/src/lib/ai-models/ai-models.component.ts
```diff
@@ -88,17 +84,11 @@ export class AiModelsComponent implements OnInit {
         tap(({ kbConfig, learningSchema }) => {
           this.kbConfigBackup = kbConfig;
           this.learningConfigurations = learningSchema;
```
mpellerin42, line 86: knowing there’s nothing left in the pipe after this step, it could be done directly in the subscribe method.

mpellerin42 (approved):

libs/common/src/lib/ai-models/summarization/summarization.component.html
```diff
@@ -30,7 +28,7 @@
           id="summary">
           @for (option of options; track option.value) {
             <pa-radio [value]="option.value">
-              {{ option | learningOption: 'summary' }}
+              {{ option.name }}
```
operramon, line 31: should it be `'kb.settings.learning-options.summary-' + option.name` as before ? or `option.name` is good enough ?

operramon (approved):

libs/common/src/lib/ai-models/summarization/summarization.component.html
ebrehault: the backend provides a proper name, no need for translation

PR #1609, opened 2024-07-04
[sc-10567] Reorganize search and widget options

mpellerin42:
- [sc-10567] Reorganize search and widget options
-  Fix widget page layout
- Prevent kb home page table to overflow


operramon (approved):

libs/common/src/lib/search-widget/search-widget.models.ts
```diff
@@ -275,7 +275,7 @@ export function getRagStrategies(ragStrategiesConfig: RagStrategiesConfig) {
 
   return { ragProperties, ragImagesProperties };
 }
-export function getNotEnoughDataMessage(config: GenerativeAnswerConfig): string {
+export function getNotEnoughDataMessage(config: WidgetConfiguration): string {
```
operramon, line 278: It seems the getNotEnoughDataMessage still doesn't work with all LLM. But I think we cannot fix it in the front.

PR #1611, opened 2024-07-04
[sc-10542] Fix budget forms

operramon:
- Allow setting an unlimited budget
- Prevent setting the budget to "0"
- Minor fix for the "metered price" field


apps/dashboard/src/app/account/billing/checkout/checkout.component.ts
```diff
@@ -479,4 +453,19 @@ export class CheckoutComponent implements OnDestroy, OnInit {
       'noreferrer',
     );
   }
+
+  modifyBudget() {
+    this.billingService.saveBudget(this.budget?.value || null).subscribe({
+      next: ({ budgetBelowTotal }) => {
+        if (budgetBelowTotal) {
+          this.toaster.warning('billing.budget_warning');
```
mpellerin42, line 461: try to be consistent in the translation keys: in most cases we're using `-` instead of `_`. Don't hesitate to update the old keys you see, replacing `_` with `-` (or with a better tree structure when appropriate).

mpellerin42 (approved):

apps/dashboard/src/app/account/billing/checkout/checkout.component.ts
operramon: ok, I'll update old keys if I find any

PR #1616, opened 2024-07-08
Display account’s creation date in manager

mpellerin42:


apps/manager-v2/src/app/manage-accounts/account-details/configuration/configuration.component.ts
```diff
@@ -37,6 +38,7 @@ export class ConfigurationComponent implements OnInit, OnDestroy {
   isSaving = false;
 
   defaultLimits?: AccountTypeDefaults;
+  created = '';
```
operramon, line 41: Is this property needed ?

operramon (approved):

apps/manager-v2/src/app/manage-accounts/account-details/configuration/configuration.component.ts
mpellerin42: No, I did a test and forgot to remove it. good catch

PR #1625, opened 2024-07-17
[sc-10660] Download resource list as CSV

operramon:


libs/common/src/assets/i18n/en.json
```diff
@@ -828,6 +828,8 @@
 	"resource.origin.related": "Related",
 	"resource.origin.url": "URL",
 	"resource.pagination.counter": "{{from}}-{{to}} of {{total}}",
+	"resource.pagination.download": "Download list",
```
ebrehault, line 831: I think we must be more specific, like "Download the entire list", because with "Download list", the user might think it will download only what is currently filtered.

ebrehault (approved):

libs/common/src/assets/i18n/en.json
operramon: Fixed

PR #1637, opened 2024-07-31
[sc-10746] Improve LLM user key management

operramon:


libs/common/src/lib/ai-models/answer-generation/answer-generation.component.ts
```diff
@@ -58,8 +59,13 @@ export class AnswerGenerationComponent extends LearningConfigurationDirective {
     }),
   });
   currentGenerativeModel?: LearningConfigurationOption;
-  hasOwnKey = false;
+  userKeyToggle = new FormControl<boolean>(false);
+  required = this.translate.stream('kb.ai-models.common.required');
```
mpellerin42, line 63: why using a stream here? with `instant` you would not need `| async` in the template

libs/common/src/lib/ai-models/answer-generation/answer-generation.component.html
```diff
@@ -54,13 +63,19 @@
...
+                        ">
                         {{ field.value.title }}
                       </pa-textarea>
                     } @else {
                       <pa-input
                         [formControlName]="field.key"
-                        [help]="field.value.info">
+                        [help]="
+                          (userKeys.required.includes(field.key) ? (required | async) + ' ' : '') +
+                          (field.value.info || '')
```
mpellerin42, line 77: if there is no `info`, we will end up with an empty `()`

mpellerin42: I just realised my mistake, those `()` are for the expression, not a string. So forgot about this comment

mpellerin42 (approved):

libs/common/src/lib/ai-models/answer-generation/answer-generation.component.ts
operramon: Fixed

libs/common/src/lib/ai-models/answer-generation/answer-generation.component.html
operramon: I just tested it. It seems fine, the parentheses are not rendered

PR #1645, opened 2024-08-05
mention cost on aitable option

ebrehault:


libs/common/src/lib/upload/upload-files/upload-files.component.html
```diff
@@ -121,17 +121,24 @@
...
+                  [value]="processingId"
+                  [help]="'upload.advanced.processing.' + processingId + '.help'">
+                  {{ 'upload.advanced.processing.' + processingId + '.label' | translate }}
                 </pa-radio>
-                @if (betaProcessings.includes(processing)) {
+                @if (betaProcessings.includes(processingId)) {
                   <nsi-badge kind="tertiary">BETA</nsi-badge>
                 }
               </div>
+              @if (processingId === 'aitable' && processing === 'aitable') {
```
mpellerin42, line 135: not sure why you need to check both `processingId` and `processing` here 🤔 
actually, I guess it's a mistake seeing `processing` doesn't exist anymore after the renaming, right?

PR #1648, opened 2024-08-06
advanced RAG pipeline

ebrehault:


libs/sdk-core/src/lib/db/search/agentic.ts
```diff
@@ -0,0 +1,439 @@
...
+ * ```typescript
+ *   const agentic = nucliaApi.knowledgeBox.createAgenticRAGPipeline({
+    START: {
+      action: {
+        type: 'predict',
+        query: 'Does the following question refers to a specific reference document? QUESTION: "{{query}}"',
+        outputs: {
+          reference: {
+            type: 'string',
+            description: 'The name of the reference document if any, and just am empty string if none.',
```
mpellerin42, line 29:
```suggestion
            description: 'The name of the reference document if any, and just an empty string if none.',
```

libs/sdk-core/src/lib/db/search/agentic.ts
```diff
@@ -0,0 +1,439 @@
...
+      } else if (step.action.type === 'ask') {
+        return this.runAskAction(stepId, step.action);
+      } else if (step.action.type === 'web') {
+        return this.runWebAction(stepId, step.action);
+      } else if (step.action.type === 'api') {
+        return this.runAPIAction(stepId, step.action);
+      } else if (step.action.type === 'user') {
+        return this.runUserAction(stepId, step.action);
+      }
+      return of(true);
```
mpellerin42, line 219: a switch would be either to read and maintain here (typescript is raising an error when a switch doesn't contain all the possible options)

libs/sdk-core/src/lib/db/search/agentic.ts
```diff
@@ -0,0 +1,439 @@
...
+        } else if (acc[key]) {
+          return acc[key];
+        } else {
+          return undefined;
+        }
+      }, obj as any);
+    }
+
+    private format(expression: string, skipMissing = true): string | undefined {
+      console.log('format', expression);
```
mpellerin42, line 261: did you mean to left this log?

mpellerin42 (approved):
Approving because as you said this code is not called anywhere, and the comments I made are quite minor. I will fix the typo and merge.
But just for you to know, I found it quite obscure, specially the different steps of formatting: the typing didn't seem aligned to me (like sending a `ChatOption` object to a method expecting a string as input…)

PR #1658, opened 2024-08-13
clean up + comments

ebrehault:


libs/sdk-core/src/lib/db/search/agentic.ts
```diff
@@ -203,103 +203,21 @@ export namespace Agentic {
...
+        case 'ask':
+          return this.runAskAction(stepId, step.action);
+        case 'web':
+          return this.runWebAction(stepId, step.action);
+        case 'api':
+          return this.runAPIAction(stepId, step.action);
+        case 'user':
+          return this.runUserAction(stepId, step.action);
+        default:
+          return of(false);
```
mpellerin42, line 220: your previous code was returning `of(true)` by default, did you change it on purpose?

libs/sdk-core/src/lib/db/search/agentic.ts
```diff
@@ -435,5 +353,118 @@ export namespace Agentic {
         }),
       );
     }
+
+    /*
```
mpellerin42, line 357: any reason why you used simple comment blocks instead of documentation mark `/**` (usually using `/**` your IDE is pre-filling the doc for you?

mpellerin42 (approved):
Way clearer!! thanks :)

libs/sdk-core/src/lib/db/search/agentic.ts
ebrehault: yes

libs/sdk-core/src/lib/db/search/agentic.ts
ebrehault: no special reason, the inspiration of the moment I guess

PR #1669, opened 2024-08-20
[sc-10891] improve method to get zones

mpellerin42:
-  Simplify getZones: Remove the check for beta testers
- Add optional cache on get subscription

### DIFF
--- a/libs/core/src/lib/api/billing.service.ts
+++ b/libs/core/src/lib/api/billing.service.ts
@@ -128,8 +128,10 @@ export class BillingService {
         this._accountSubscriptionCache.next(subscription);
         return subscription;
       }),
-      catchError(() => {
-        if (fromCache) {
+      catchError((error) => {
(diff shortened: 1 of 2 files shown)


libs/core/src/lib/api/billing.service.ts
```diff
@@ -95,17 +113,27 @@ export class BillingService {
...
+          subscription = data as AccountSubscription;
         }
+
+        this._accountSubscriptionInit = true;
+        this._accountSubscriptionCache.next(subscription);
+        return subscription;
+      }),
+      catchError(() => {
+        if (fromCache) {
+          this._accountSubscriptionInit = true;
```
ebrehault, line 133: so if we get an error on the first call, we will return `null` forever?

ebrehault (approved):

libs/core/src/lib/api/billing.service.ts
mpellerin42: yes

operramon (approved):

PR #1671, opened 2024-08-20
Ericbrehault/sc 10431/replace status json and announce json with

ebrehault:


apps/nucliadb-admin/src/app/app.component.ts
```diff
@@ -87,32 +82,4 @@ export class AppComponent implements OnInit, OnDestroy {
...
-    this.checkMessages(true);
-  }
-
-  private displayAnnounce() {
-    this.checkMessages(false);
-  }
-
-  private checkMessages(alert: boolean) {
-    this.tracking
-      .getStatusMessage(alert)
```
operramon, line 101: can `getStatusMessage` method be removed too ?

PR #1683, opened 2024-08-29
[sc-10921] Reset the question when leaving the search/widget page

mpellerin42:


ebrehault (approved):

libs/common/src/lib/search-widget/search-widget.service.ts
```diff
@@ -213,6 +213,10 @@ export class SearchWidgetService {
     );
   }
 
+  resetSearchQuery() {
+    this.currentQuery = '';
```
operramon, line 217: should `this.currentFilters` also be reset ?

operramon (approved):

PR #1684, opened 2024-08-29
new dev environment

ebrehault:


libs/common/src/lib/topbar/topbar.component.ts
```diff
@@ -17,7 +17,7 @@ export class TopbarComponent {
   userInfo = this.userService.userInfo;
   account = this.sdk.currentAccount;
   kb = this.sdk.currentKb;
-  isStage = location.hostname === 'stashify.cloud';
+  isStage = location.hostname === 'stashify.cloud' || location.hostname === 'gcp-global-dev-1.nuclia.io';
```
mpellerin42, line 20: why not using `isStageOrDev` from feature-flag service here? I feel like having `gcp-global-dev-1.nuclia.io` duplicated in several places is a not a good idea

mpellerin42 (approved):

PR #1726, opened 2024-09-10
[sc-9890] Improve text resource upload

operramon:


libs/common/src/assets/i18n/fr.json
```diff
@@ -1434,8 +1434,8 @@
 	"upload.error": "Échec du téléchargement",
 	"upload.files": "Téléchargement de fichiers",
 	"upload.folder": "Télécharger un dossier",
-	"upload.invalid-csv-labels": "Fichier CSV non valide. Le format de la colonne <strong>Étiquettes</strong> est incorrect.",
-	"upload.invalid_csv": "Fichier CSV non valide. Veuillez vous assurer que chaque ligne contient exactement {{num}} colonnes.",
+	"upload.invalid-csv-labels": "CSV non valide: le format de la colonne <strong>Libellés</strong> est incorrect.",
```
mpellerin42, line 1437: I think we used `étiquettes` instead of `libellés` on the other translations of labels. Can you check so we keep consistency accross the different translations?

mpellerin42 (approved):

PR #1728, opened 2024-09-11
expose API errors

ebrehault:


libs/search-widget/src/core/api.ts
```diff
@@ -447,3 +448,13 @@ export function downloadDump() {
...
     .subscribe((data) => downloadAsJSON(data));
 }
+
+export function getApiErrors() {
+  return merge(searchError, suggestionError, chatError).pipe(
+    filter((error) => !!error?.status),
+    map((error) => ({
+      status: error?.status,
+      detail: error?.detail,
+    })),
```
mpellerin42, line 458: the method is named `getApiErrors` but from what I can see you can get only one error at a time. I don't know the use case behind so that might be what we want, just wondering.

mpellerin42 (approved):

operramon (approved):

libs/search-widget/src/core/api.ts
ebrehault: yes that's what we want

PR #1733, opened 2024-09-13
[sc-10707] show existing tasks in the dashboard

mpellerin42:
## What have been done:
- Add support for Task management in the SDK:
  - new `taskManager` property in `IWritableKnowledgeBox` implemented by a new `TaskManager` class
  - creation of all the models and methods based on the first implementation of the API, they are fully documented in the code.
- Show existing text blocks labeler tasks in the dashboard:
  - creation of UI models and mapper from backend models to UI models
  - creation of `TasksAutomationService` doing the interface between the UI and the SDK
  - using the new service to initialize the task list 
  - Removing the `Create task` button and adding a disclaimer on the task list page explaining the ability to create a new task from the dashboard will come soon
  - Keeping only the text blocks labeler task list visible on tasks list page
  - Add a menu on the task list item (not finished yet)
  - Show fake tasks on the list with a mock

## Left to do:
- Mapper for the parameters (in `task-automation.models.ts`: I wasn't sure how the parameters would be structured, I was waiting for a real backend sample from Ramon to finalize this method
- The [API documentation](https://docs.stashify.cloud/docs/nua-api/#tag/Task) didn't change after Ramon merged his last big PR, so the models and methods might need to be updated accordingly to the last changes on the API
- Display the labels on the task list
- Not sure if we will be able to keep the `Activity` column on this first iteration, depending on the API capabilities you might need to remove it
- Finish the task list menu which should allow to stop/start a task (to be confirmed with Ramon as the 1st version of the API didn't allow to restart a task that was stopped, but we discussed it and that should have been fixed), delete a task.

## Next iteration(s)
- Vincent designs contained a page to view a tasks details, this page wasn't implemented (as it would be easier to implement it once we have real data) but could be done on a 2nd iteration. Once implemented, a menu to access it should be added.
- The Task creation form was done in a generic yet static way, the general look and feel is the right one but the API to create tasks is quite complex (containing some kinds of schemas) so you might need to review the way the UI is built to make it work properly with the API


libs/common/src/assets/i18n/fr.json
```diff
@@ -1444,6 +1445,7 @@
 	"tasks-automation.table.header.ner-family": "Famille NER",
 	"tasks-automation.table.header.status": "Statut",
 	"tasks-automation.table.processed": "{{count}} traités",
+	"tasks-automation.task-creation-info": "Vous pourrez bientôt créer vos tâches directement à partir de cette page. En attendant, veuillez suivre la documentation pour les créer à l'aide de l'API.",
```
ebrehault, line 1448:
```suggestion
	"tasks-automation.task-creation-info": "Vous pourrez bientôt créer vos tâches directement à partir de cette page. En attendant, veuillez suivre la documentation pour les créer à l‘aide de l‘API.",
```

operramon (approved):

PR #1756, opened 2024-10-03
Display attached images in results

ebrehault:


libs/search-widget/src/components/image/Image.scss
```diff
@@ -0,0 +1,26 @@
...
+    bottom: 0;
+    display: flex;
+    left: 0;
+    justify-content: center;
+    padding: var(--main-padding-top) var(--main-padding-side) 0;
+    position: fixed;
+    right: 0;
+    top: 0;
+    z-index: var(--z-index-modal);
+    background-color: var(--);
```
operramon, line 19: is this line correct ?

libs/sdk-core/src/lib/db/search/utils.ts
```diff
@@ -99,4 +99,5 @@ export interface WidgetFeatures {
   orFilterLogic?: boolean;
   noChatHistory?: boolean;
   showHidden?: boolean;
+  showAttachedImages?: boolean;
```
operramon, line 102: at the moment, this option shouldn't be visible in the dashboard ?

operramon (approved):

PR #1768, opened 2024-10-09
support audit_metadata in widget

ebrehault:


libs/search-widget/src/widgets/search-widget/SearchBar.svelte
```diff
@@ -85,6 +85,7 @@
   export let json_schema = '';
   export let vectorset = '';
   export let chat_placeholder = '';
+  export let audit_metadata = '';
```
operramon, line 88: should also be added to popup and chat widgets ?

libs/common/src/lib/search-widget/search-widget.service.ts
```diff
@@ -194,6 +194,7 @@ export class SearchWidgetService {
 
         let baseSnippet = `<${tagName}${theme}\n  knowledgebox="${kb.id}"`;
         baseSnippet += `\n  ${zone}${features}${prompt}${systemPrompt}${rephrasePrompt}${ragProperties}${ragImagesProperties}${placeholder}${chatPlaceholder}${notEnoughDataMessage}${askToResource}${maxTokens}${maxParagraphs}${queryPrepend}${generativeModel}${vectorset}${filters}${preselectedFilters}${privateDetails}${backend}${jsonSchema}`;
+        baseSnippet += `\n  audit_metadata="{\\"config\\":\\"${currentConfig.id}\\"}"`;
```
operramon, line 197: is this line added on purpose ?

operramon (approved):

libs/search-widget/src/widgets/search-widget/SearchBar.svelte
ebrehault: right, fixed!

libs/common/src/lib/search-widget/search-widget.service.ts
ebrehault: Yes, by default we put the search config id in the audit metadata, and the users are free to add more manually after pasting the snippet if they want to.

PR #1839, opened 2024-11-13
add lang option

ebrehault:


libs/common/src/lib/search-widget/search-widget.service.ts
```diff
@@ -145,6 +146,7 @@ export class SearchWidgetService {
     const features = getFeatures(currentConfig, widgetOptions);
     const placeholder = getPlaceholder(widgetOptions);
     const chatPlaceholder = getChatPlaceholder(widgetOptions);
+    const lang = getLang(widgetOptions);
```
operramon, line 149: This `lang` option is not passed to the preview widget. Is it expected ?

operramon (approved):

PR #1842, opened 2024-11-15
[sc-11224] Support content safety agents

operramon:


ebrehault (approved):
This task is not ready yet (and it not even be `llm-align`). So let's add a specific feature-flag on it, so we can merge the code.

PR #1908, opened 2025-01-10
Ericbrehault/sc 11388/fix copy dsiclaimer

ebrehault:


libs/search-widget/src/widgets/chat-widget/ChatWidget.svelte
```diff
@@ -132,6 +132,7 @@
       (acc, current) => ({ ...acc, [current as keyof WidgetFeatures]: true }),
       {},
     );
+    console.log('features', _features);
```
operramon, line 135: is it on purpose ?

operramon (approved):

libs/search-widget/src/widgets/chat-widget/ChatWidget.svelte
ebrehault: nope :)

PR #1909, opened 2025-01-10
[sc-11314] Add option to collapse text blocks

operramon:


ebrehault (approved):
In my opinion, if the `collapseTextBlocks` option is on, the user will alredy have to click once to see the details of a given result, so we do not want them to need to click anything else.
So the we should not collapse paragraphs and we should directly display all results:

image

PR #1911, opened 2025-01-10
remove content type filters in agents

ebrehault:


libs/common/src/lib/tasks-automation/task-forms/task-form.component.ts
```diff
@@ -305,9 +265,7 @@ export class TaskFormComponent implements OnInit, OnDestroy {
   }
 
   onToggleFilter() {
-    this.allResourceTypesSelected = this.resourceTypeFilters.every((option) => option.selected);
-    this.allFieldTypesSelected = this.fieldTypeFilters.every((option) => option.selected);
```
operramon, line 309: A small thing, maybe this line (309) shouldn't be removed

operramon (approved):

PR #1945, opened 2025-02-05
load and display logs in dashboard

ebrehault:


apps/dashboard/src/app/activity/activity-download.component.ts
```diff
@@ -102,4 +182,12 @@ export class ActivityDownloadComponent {
     const date = new Date(parseInt(year), parseInt(month) - 1, 2);
     return format(date, 'MMMM yyyy');
   }
+
+  toggle(url?: string) {
+    if (url) {
+      from(fetch(url).then((res) => res.text())).subscribe((ndjson) => {
+        console.log(ndjson);
```
operramon, line 189: is it on purpose ?

apps/dashboard/src/app/activity/activity-download.component.html
```diff
@@ -29,24 +29,39 @@ <h1>{{ 'navbar.activity' | translate }}</h1>
     </pa-tabs>
     <div class="download-list">
       @for (item of activity[selectedActivityTab] | async; track item.month) {
```
operramon, line 31: I think I wrote this line wrong.
Could we change it to `; track selectedActivityTab + item.month` ?

operramon (approved):

apps/dashboard/src/app/activity/activity-download.component.ts
ebrehault: no, useless code

PR #1989, opened 2025-03-13
support max_pages_to_merge

ebrehault:


libs/common/src/lib/ai-models/extraction/extraction-modal/llm-configuration.component.ts
```diff
@@ -77,6 +82,7 @@ export class LLMConfigurationComponent implements OnDestroy, OnInit {
         this.valueChange.emit({
           llm: values.customLLM ? this.getLLMConfig(values.llm.generative_model) : undefined,
           merge_pages: values.merge_pages,
+          max_pages_to_merge: values.max_pages_to_merge,
```
operramon, line 85: when `merge_pages` is disabled,  should `undefined` be sent in `max_pages_to_merge` ?

operramon (approved):

libs/common/src/lib/ai-models/extraction/extraction-modal/llm-configuration.component.ts
ebrehault: yes, that's probably safer, let me fix it

PR #1994, opened 2025-03-18
[sc-11803] Add data augmentation testing tab

operramon:


libs/common/src/lib/tasks-automation/task-details/task-testing/task-testing.component.ts
```diff
@@ -0,0 +1,146 @@
...
+    this.results = undefined;
+    this.tokens = undefined;
+    this.cdr.markForCheck();
+    this.sdk.currentKb
+      .pipe(
+        take(1),
+        switchMap((kb) =>
+          forkJoin(
+            this.selectedIds.map((id) => {
+              const resource = new Resource(
```
ebrehault, line 92: The method `getResourceFromData()` can do that for you.
But no need to change, that's fine.

ebrehault (approved):

PR #2017, opened 2025-03-28
DA agents: support AND/OR on labels and keywords

ebrehault:


libs/common/src/lib/tasks-automation/task-details/task-settings/task-settings.component.html
```diff
@@ -30,6 +30,11 @@
         <div class="setting">
           <label>{{ 'tasks-automation.form.filter.search-in.title-or-content' | translate }}</label>
           <div>{{ filter.contains?.[0] || '-' }}</div>
```
operramon, line 32: A minor thing, only the first keyword is displayed

operramon (approved):

PR #2030, opened 2025-04-04
Ericbrehault/sc 12134/white labelled manager favicon env config

ebrehault:


libs/common/src/lib/navbar/navbar.component.html
```diff
@@ -264,7 +264,7 @@
               <span class="nav-link-title">{{ 'navbar.nua-keys' | translate }}</span>
             </a>
           </div>
-          @if (isBillingEnabled | async) {
+          @if ((isBillingEnabled | async) && !noStripe) {
```
operramon, line 267: Should the "Payment & Subscription" link in the topbar be also hidden ?

operramon (approved):

libs/common/src/lib/navbar/navbar.component.html
ebrehault: fixed

PR #2031, opened 2025-04-04
Create / Delete retrieval agents from account

mpellerin42:


libs/common/src/lib/navbar/navbar.component.html
```diff
@@ -264,7 +264,17 @@
               <span class="nav-link-title">{{ 'navbar.nua-keys' | translate }}</span>
             </a>
           </div>
-          @if ((isBillingEnabled | async) && !noStripe) {
```
operramon, line 267: Hello,
a minor thing: should this line be changed ?

operramon (approved):

ebrehault (approved):

libs/common/src/lib/navbar/navbar.component.html
mpellerin42: there was a conflict there, I thought I kept the main version but I probably did it wrong. I'll fix it to be like it is currently on stage

PR #2081, opened 2025-04-28
[sc-12121] retrieval agent workflow editor

mpellerin42:


apps/dashboard/src/app/retrieval-agent/agent-dashboard/workflow/workflow.state.ts
```diff
@@ -0,0 +1,145 @@
...
+export function updateNode(
+  id: string,
+  agent: PreprocessAgent | ContextAgent | PostprocessAgent,
+  nodeConfig: NodeConfig,
+) {
+  const node = getNode(id);
+  if (node) {
+    nodes.update((items) => ({ ...items, [id]: { ...node, agent, nodeConfig } }));
+    node.nodeRef.setInput('config', nodeConfig);
+  }
```
ebrehault, line 79: (Just a code style thing)
Here you use `update` to assign the signal value after the logic happened (a bit like we would use `set`).
In my opinion, it is more natural to wrap the full logic in the `update` call, like:
```
nodes.update((items) => {
  const node = items[id];
  if (node) {
    items[id] = {...node, agent, nodeConfig };
    node.nodeRef.setInput('config', nodeConfig);
  }
  return items;
});
```

That would be similar to what we do with the store writers in the widget for example.

But no big deal, the code works like that too, no need to change.

apps/dashboard/src/app/retrieval-agent/agent-dashboard/workflow/workflow.service.ts
```diff
@@ -121,6 +138,85 @@ export class WorkflowService {
   sideBarOpen = computed(() => this._sideBarOpen());
   activeSideBar = computed(() => this._activeSideBar());
 
+  /**
+   * Initialise the workflow with the agents saved in current Arag, update the workflow when current Arag changes.
+   */
+  initAndUpdateWorkflow(root: WorkflowRoot) {
```
ebrehault, line 144: I guess it is under refactoring and not entirely migrated to the state approach.
The objective is to have a clean separation here:
- when reading the current arag data from the backend, we should just update the state
- when the state changes:
  - some `effect` must create the UI components
  - some `effect` must make the backend calls if needed and if possible (for example, not if a THEN is missing or things like that)

ebrehault (approved):

apps/dashboard/src/app/retrieval-agent/agent-dashboard/workflow/workflow.state.ts
mpellerin42: I see what you mean. I did it that way because I don't really like the idea of calling a signal (`setInput` is internally a signal) inside another signal’s update…

mpellerin42:
Replaced by https://github.com/nuclia/frontend/pull/2084

PR #2084, opened 2025-04-29
[sc-12121] arag workflow editor with state

mpellerin42:
## Done:
- workflow and its sidebar now managed with a state
- Saving in the backend now managed in an effect
- Update MCP driver model (now split in two: MCP SSE and MCP STDIO)
- Ability to save child nodes for fallback, then, else

## In progress:
- children management in a recursive way
- I left a lot of console.debug in purpose


apps/dashboard/src/app/retrieval-agent/agent-dashboard/agent-dashboard.component.scss
```diff
@@ -2,6 +2,8 @@
 @use 'agent-dashboard.tokens' as dashboard;
 
 .agent-dashboard {
+  position: relative;
```
operramon, line 5: This rule seems to make the sidebar div always visible (when scrolling to the right), event when the sidebar is closed. I don't know if it's a bug or not.

operramon (approved):

apps/dashboard/src/app/retrieval-agent/agent-dashboard/agent-dashboard.component.scss
mpellerin42: definitely a bug yes, good catch! I added this property when I tried to add a mask when loading the workflow, but it was giving a blinking effect worst than not having a mask, so I removed the mask and forgot this property…
I'll remove it now so I don't forget, thanks!!

apps/dashboard/src/app/retrieval-agent/agent-dashboard/workflow/nodes/ask/ask-form.component.html
```diff
@@ -48,12 +48,12 @@
         resizable="true"></pa-textarea>
     </div>
     <div class="form-entry">
-      <label for="keyword_custom_prompt">
-        {{ 'retrieval-agents.workflow.node-types.ask.form.keyword_custom_prompt' | translate }}
+      <label for="keywords_custom_prompt">
+        {{ 'retrieval-agents.workflow.node-types.ask.form.keywords_custom_prompt' | translate }}
       </label>
       <pa-textarea
```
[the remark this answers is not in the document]
mpellerin42: oh I didn't know! Is there any documentation page on sistema-demo?

ebrehault (approved):

PR #2095, opened 2025-05-09
[sc-12121] ARAG editor for all agents types

mpellerin42:
### Done
- Add optional temporary rules to all agents
- Setup node and form for all agents
- Fix `deepEqual` function to consider `null` and `undefined` properties as equivalent
- Add google driver form
- Manage cases where there is no driver/source available in all agents forms


apps/dashboard/src/app/retrieval-agent/agent-dashboard/workflow/nodes/external/external-form.component.html
```diff
@@ -0,0 +1,112 @@
...
+            resizable
+            help="retrieval-agents.workflow.node-types.external.form.call_obj.help"
+            modalTitle="retrieval-agents.workflow.node-types.external.form.call_obj.label"></nsi-expandable-textarea>
+        </div>
+      }
+    }
+
+    <app-headers-field
+      [form]="configForm"
+      [config]="externalConfig"></app-headers-field>
```
operramon, line 106: A minor thing:
If a "External call" node with headers is created,
then when opening it in the sidebar, the displayed headers do not seem to be the same.

operramon (approved):

PR #2098, opened 2025-05-12
[sc-12121] arag editor improvements

mpellerin42:
- Prevent node links to cross each others
- Highlight unsaved nodes

### DIFF
--- a/apps/dashboard/src/app/retrieval-agent/agent-dashboard/agent-dashboard.component.scss
+++ b/apps/dashboard/src/app/retrieval-agent/agent-dashboard/agent-dashboard.component.scss
@@ -29,17 +29,20 @@
     width: 100%;
     transition: width $transition-slide;
 
+    $gap: rhythm(3);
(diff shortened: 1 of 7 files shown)


apps/dashboard/src/app/retrieval-agent/agent-dashboard/agent-dashboard.component.scss
```diff
@@ -33,8 +33,14 @@
       display: flex;
       flex-direction: column;
       gap: rhythm(3);
-      height: fit-content;
       padding: rhythm(8);
+
+      .column-section {
```
operramon, line 38: It seems that "column sections" might overlap each other when there are many nodes. I don't know if it's a problem.

mpellerin42: definitely a problem yes, good catch !

PR #2108, opened 2025-05-15
[sc-12305] Fix search configurations in search page

operramon:


libs/common/src/lib/search-widget/search-configuration/search-configuration.component.ts
```diff
@@ -45,6 +47,13 @@ import { PaTogglesModule } from '../../../../../pastanaga-angular/projects/pasta
 
 const NUCLIA_SEMANTIC_MODELS = ['ENGLISH', 'MULTILINGUAL', 'MULTILINGUAL_ALPHA'];
 
+@Pipe({ name: 'isTypedConfig' })
```
ebrehault, line 50: Using a pipe is an excellent idea!

ebrehault (approved):

PR #2159, opened 2025-06-04
apply widget config loaded from backend

ebrehault:


operramon (approved):

libs/search-widget/src/core/utils.ts
```diff
@@ -532,7 +532,8 @@ export function loadWidgetConfig(id: string, options: NucliaOptions) {
         console.error(`Search configuration not found: "${widget.searchConfigId}"`);
         return of({});
       }
-      return getWidgetParameters(searchConfig, widget.widgetConfig);
+      const params = getWidgetParameters(searchConfig, widget.widgetConfig);
+      return Object.fromEntries(Object.entries(params).filter(([_, value]) => value !== null && value !== ''));
```
operramon, line 536: one question,
why empty strings are filtered out and not `undefined` ?

ebrehault: To preserve the default values set in the widget code.
We do:
`let config = $state(new Props());`
to initialize the config with default values (like `backend = 'https://nuclia.cloud/api'`), and if we get `backend: ''` from the backend config, we would lose it.

PR #2185, opened 2025-06-13
[sc-12443] widgets in pure css

mpellerin42:
- Rename all scss files to css
- [@keyframes cannot be nested in css](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting/Nesting_at-rules#at-rules_that_can_be_nested) so move @keyframes out of nesting
- Fix Svelte `:global` block usage in Markdown component


ebrehault (approved):
all good, it does not break the white-labelling thing

libs/search-widget/src/common/indicators/TimeIndicator.scss
operramon, line 1: should `TimeIndicator.scss` be renamed to ".css" ?

libs/search-widget/src/components/viewer/renderers/ImageRenderer.svelte
```diff
@@ -1,14 +1,14 @@
...
 <div class="sw-image-renderer">
-  <img src={$imageUrl} alt={$fieldData?.value?.file?.filename || ''} />
+  <img
+    src={$imageUrl}
+    alt={$fieldData?.value?.file?.filename || ''} />
 </div>
 
-<style
-  lang="scss"
-  src="./ImageRenderer.scss"></style>
```
operramon, line 14: A minor thing, it seems `ImageRenderer.scss` hasn't been renamed

operramon (approved):

PR #2207, opened 2025-06-25
[sc-12476] Show local time in activity log

operramon:


apps/dashboard/src/app/activity/log-table.component.html
```diff
@@ -50,14 +50,16 @@
       }
       @for (row of displayedRows; track row.id) {
         <pa-table-row>
-          @if (displayedHeaders.includes('Date')) {
+          @if (displayedHeaders.includes(dateColumn)) {
             <pa-table-cell>
-              <div class="body-s">
+              <div
+                class="body-s"
+                [paTooltip]="(row.date + 'Z' | date: 'MMM d, h:mm a') + ' (' + ('activity.local-time' | translate) + ')'">
```
mpellerin42, line 57: I understand you add the `Z` because the backend doesn't return the timezone in the timestamp, but it would be better if this fix was done on their side if possible: if they do the fix at some point this code won't work anymore otherwise.

mpellerin42 (approved):

ebrehault (approved):

PR #2277, opened 2025-09-03
Use the product id instead of the product code

sunbit:
this optionallity is to avoid breaking things can be set to use only aws_product_id later

### DIFF
--- a/apps/dashboard/src/app/account/billing/subscriptions/aws-subscription/aws-subscription.component.ts
+++ b/apps/dashboard/src/app/account/billing/subscriptions/aws-subscription/aws-subscription.component.ts
@@ -17,7 +17,7 @@ export class AwsSubscriptionComponent {
     .pipe(
       map((subscription) => {
         const productId = subscription?.aws_product_id || subscription?.aws_product_code;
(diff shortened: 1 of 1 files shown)


apps/dashboard/src/app/account/billing/subscriptions/aws-subscription/aws-subscription.component.ts
```diff
@@ -15,10 +15,10 @@ export class AwsSubscriptionComponent {
   awsUrl = this.billing
     .getAwsSubscription()
     .pipe(
-      map(
-        (subscription) =>
-          `https://console.aws.amazon.com/marketplace/home#/subscriptions/${subscription?.aws_product_code || ''}`,
-      ),
+      map((subscription) => {
+        const productId = subscription?.aws_product_id || subscription?.aws_product_code;
+        return `https://console.aws.amazon.com/marketplace/home#/subscriptions/${productId}`;
```
ebrehault, line 20: productId can be undefined

```suggestion
        return `https://console.aws.amazon.com/marketplace/home#/subscriptions/${productId || ''}`;
```

ebrehault (approved):

PR #2745, opened 2026-04-14
Updated texts, fixed styles

ykulynych:


libs/common/src/lib/metrics/metrics-filters/metrics-filters.component.ts
```diff
@@ -186,22 +191,24 @@ export class MetricsFiltersComponent {
     return this.service.getOperationsForColumn(key, this.filterColumns());
   }
 
+  getMonthLastDay(): string {
+    const month = this.selectedMonth();
+    if (!month) return '';
+    const [year, mo] = month.split('-');
+    const lastDay = new Date(Number(year), Number(mo), 0).getDate();
```
operramon: By the way, the `date-fns` is included in package.json.
Sometimes I find it useful to do date calculations.
But it's not mandatory to use it, it's just an additional option we have.

operramon (approved):

PR #2753, opened 2026-04-16
[sc-14271]: Show errors clearly in debug mode

ykulynych:


libs/search-widget/src/core/stores/effects.ts
```diff
@@ -575,9 +580,10 @@ export function askQuestion(
         if (!hasError) {
           // error is set only once
           hasError = true;
-          const error = translateInstant(
+          const translatedError = translateInstant(
             result.status === -2 ? getNotEngoughDataMessage() : messages[`${result.status}`] || 'error.search',
           );
+          const error = isDebugMode && result.detail ? result.detail : translatedError;
```
operramon, line 586: While testing I noticed a minor thing:
If a wrong parameter is sent to the `/ask` endpoint, the backend returns code 412,
and the widget displays the message: `[object Object]`.
Maybe `result.detail` is an object in this case

operramon (approved):

libs/search-widget/src/core/stores/effects.ts
ykulynych: Ah, good spot. I’ll fix that.

PR #2755, opened 2026-04-21
[sc-14396] Enabled toggle to enforce security during kb creation

ykulynych:


libs/common/src/lib/kb-creation/kb-creation.component.html
operramon, line 1: While testing I noticed one small thing:
when creating a Knowledge Box from the onboarding pages, the "enforce_security" is set to false.
Should this specific case also be addressed ?

operramon (approved):

libs/common/src/lib/kb-creation/kb-creation.component.html
```diff
@@ -77,6 +77,14 @@ <h1 class="page-title">{{ 'account.kb.creation-form.title' | translate }}</h1>
         </div>
       </nsi-two-columns-configuration-item>
 
+      <nsi-two-columns-configuration-item itemTitle="account.kb.creation-form.enforce-security.title">
```
ebrehault: Sorry I was not clear: I don't want it as an option during the creation flow (we want to keep this flow simple).
I just want the parameter to be always set to true.

(And after creation, the user can change it if they want by going to the Settings page)

ebrehault (changes requested):

libs/common/src/lib/kb-creation/kb-creation.component.html
ykulynych: Understood, removed toggle

libs/common/src/lib/kb-creation/kb-creation.component.html
ykulynych: I think so, fixed it

ebrehault (approved):

PR #2768, opened 2026-04-24
allow to preset the workflow

ebrehault:


libs/common/src/lib/onboarding/onboarding.component.ts
```diff
@@ -72,6 +72,9 @@ export class OnboardingComponent {
       .subscribe((account) => {
         this.account = account;
         this.creatingAccount = false;
+        if (this.account.workflow === 'cowork') {
```
operramon, line 75: At this point in time,  the workflow is still not set. It is set after step 2.

libs/common/src/lib/onboarding/onboarding.component.ts
```diff
@@ -82,7 +85,7 @@ export class OnboardingComponent {
   }
 
   storeWorkflowAndGoNext(workflow: WorkflowType) {
-    this.kbName = 'Cowork';
+    this.kbName = 'ContextBox';
```
operramon: I think I made a mistake in my previous PR.
Should this line be executed only if "workflow" is "cowork" ?

ebrehault: oh, right, let me fix it

libs/common/src/lib/onboarding/onboarding.component.ts
ebrehault: Carles made a change on signup flow, we can now set the workflow at the very beginning (in the form on progress.com).
So my change is about that: if the workflow is already set, then we skip the step 2.

operramon (approved):

libs/common/src/lib/onboarding/onboarding.component.ts
operramon: oh, I understand.
will `workflow` be set automatically to "cowork" even if we don't pass this parameter when creating the account ?

PR #2771, opened 2026-04-27
Allow for shorten token time with infinite loop refresh

sunbit:
The frontend was refreshing tokens in a tight infinite loop whenever the access token lifetime was shorter than 6 hours, because the refresh timer was hardcoded to either "now" or "6 hours" with no middle ground. This fixes the scheduling logic to refresh 1 minute before the token actually expires, making it work correctly with any lifetime. 

This unblocks reducing first-party app tokens from 120h to 1 hour.


sunbit:
sc-14460

ebrehault (approved):

PR #2814, opened 2026-05-12
Georgekulynych/sc 14681/do not allow video and audio files in cowork

ykulynych:


operramon (approved):

libs/common/src/assets/i18n/en.json
```diff
@@ -2323,6 +2323,7 @@
 	"simple.mcp.how-to": "How to use your Content Box via MCP?",
 	"simple.mcp.title": "MCP server URL",
 	"simple.no-questions": "You haven't asked any questions yet.",
+	"simple.no-video-audio": "Video and audio files are not supported",
```
ebrehault, line 2326: Actually they are supported but not allowed in this tier. Let's be explicit: "Video and audio files are not allowed in your current subscription"

ebrehault (approved):

PR #2831, opened 2026-05-18
exhaust catalog pages

ebrehault:


apps/dashboard/src/app/knowledge-box/simple/simple-kb/simple-kb.service.ts
operramon: I seems it can now be replaced with "this.catalog(..."

operramon (approved):

PR #2843, opened 2026-05-20
[sc-14539] Added zone visibility to account, origin and privacy field…

ykulynych:
…s to zones

### DIFF
--- a/libs/core/src/lib/api/zone.service.ts
+++ b/libs/core/src/lib/api/zone.service.ts
@@ -1,11 +1,11 @@
 import { Injectable } from '@angular/core';
 import { Zone } from '../models';
-import { catchError, map, Observable, of, shareReplay, switchMap } from 'rxjs';
+import { catchError, map, Observable, of, shareReplay, switchMap, take } from 'rxjs';
 import { FeatureFlagService } from '../analytics/feature-flag.service';
 import { SDKService } from './sdk.service';
-import { take } from 'rxjs/operators';
 import { BillingService } from './billing.service';
 import { AwsAccountSubscription } from '../models/billing.model';
+import { setZoneInRegionalUrl } from '@nuclia/core';
 
 const ZONES = 'zones';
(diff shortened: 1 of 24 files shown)


libs/core/src/lib/api/zone.service.ts
```diff
@@ -20,9 +20,12 @@ export class ZoneService {
...
-  // state, so that part must stay live per subscriber.
-  private readonly rawZones$: Observable<Zone[]> = this.sdk.nuclia.rest.get<Zone[]>(`/${ZONES}`).pipe(shareReplay(1));
+  // Account-scoped: zones are fetched per account so the user only sees zones their account
+  // can actually use. Re-fetches automatically when the active account changes.
+  private readonly rawZones$: Observable<Zone[]> = this.sdk.currentAccount.pipe(
+    switchMap((account) => this.sdk.nuclia.rest.get<Zone[]>(`/account/${account.slug}/${ZONES}`)),
+    shareReplay(1),
+  );
 
   getZones(): Observable<Zone[]> {
```
operramon, line 30: A minor thing:
while testing, I noticed that getZones() now returns a "uncomplete" stream instead of a "complete" one.
It seems that this subtle change is causing small issues at least in these 3 places:

-https://github.com/nuclia/frontend/blob/fccf5e9e6286571ab19c153f706a30f80364217c/libs/common/src/lib/account/account-nua/client-dialog/client-dialog.component.ts#L87

- https://github.com/nuclia/frontend/blob/fccf5e9e6286571ab19c153f706a30f80364217c/libs/common/src/lib/account/account-models/custom-models/custom-models.component.ts#L66

- https://github.com/nuclia/frontend/blob/fccf5e9e6286571ab19c153f706a30f80364217c/libs/common/src/lib/account/account-models/account-models.component.ts#L272

It seems like these "forkjoin" don't emit a value anymore.

operramon (approved):

PR #2850, opened 2026-05-22
[sc-14813] Improve history privacy

operramon:
- It's better to use the user Id in `security.groups` instead of the user name, because user id is immutable and unique.

### DIFF
--- a/apps/dashboard/src/app/knowledge-box/simple/simple-kb/simple-kb.service.ts
+++ b/apps/dashboard/src/app/knowledge-box/simple/simple-kb/simple-kb.service.ts
@@ -65,10 +65,8 @@ export class SimpleKBService {
     failed: 0,
   });
 
-  userId = this.sdk.nuclia.auth.getAuthenticatedUser().pipe(
-    map((user) => user.id),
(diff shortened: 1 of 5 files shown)


apps/dashboard/src/app/knowledge-box/simple/simple-kb/simple-kb.service.ts
```diff
@@ -67,7 +65,10 @@ export class SimpleKBService {
     failed: 0,
   });
 
-  userName = this.userService.userPrefs.pipe(map((userPrefs) => userPrefs?.name || ''));
+  userId = this.sdk.nuclia.auth.getAuthenticatedUser().pipe(
```
ykulynych: It looks that we need only userId here, can't we take it this way?
```
 const user = nuclia.auth.getJWTUser();
 const userId = user?.sub; //  <- user ID
```

operramon: ah yes, I'll change it

ykulynych (approved):

PR #2865, opened 2026-05-29
[sc-14817] Added settings page for contextbox

ykulynych:


ebrehault (approved):

libs/common/src/lib/topbar/topbar.component.ts
```diff
@@ -102,6 +102,13 @@ export class TopbarComponent {
       this.navigationService.kbUrl.pipe(take(1)),
       this.navigationService.homeUrl.pipe(take(1)),
     ]).subscribe(([isCowork, kbUrl, homeUrl]) => {
+      if (this.inDashboard && !isCowork) {
+        // The logo should always take dashboard users back to the simple home UI.
```
operramon: This fragment of code seems strange, maybe it's a LLM hallucination

libs/core/src/lib/services/navigation.service.ts
```diff
@@ -216,7 +216,22 @@ export class NavigationService {
     this.router.navigate([this.getAccountSelectUrl()]);
   }
 
-  setSimpleMode(value: boolean) {
-    this.simpleMode.next(value);
+  private modeManuallyOverridden = false;
```
operramon, line 219: I don't know why the llm added the `modeManuallyOverridden` variable,
is it useful ?

operramon (approved):

libs/common/src/lib/topbar/topbar.component.ts
ykulynych: No it was added intentionally, but looks a bit wired to me also. Reworked it.

libs/core/src/lib/services/navigation.service.ts
ykulynych: It was added to fix race condition when you toggle simple mode to advanced, but setAccountGuard toggle it back.
PS. Don't like this toggle, needed to fix few issues related to it.

PR #2872, opened 2026-06-02
[sc-14822] Added more options to reasoning effort picker

ykulynych:


libs/common/src/lib/search-widget/search-configuration/generative-answer-form/generative-answer-form.component.html
operramon, line 1: One question, I remember there is a component called `stf-reasoning-config`,
should it also be modified ?

ykulynych: I thought that they were different things, because configs contained different options (one with 'minimal', another without it), but i need to check

ykulynych: Added the new options to stf-reasoning-config also.

operramon (approved):

PR #2883, opened 2026-06-04
[sc-14900] Fix the model restrictions on JSON output

operramon:


libs/common/src/lib/search-widget/search-configuration/results-display-form/results-display-form.component.ts
```diff
@@ -176,6 +159,22 @@ export class ResultsDisplayFormComponent implements OnInit, OnDestroy {
...
+    effect(() => {
+      if (this.generativeModel()) {
+        const structredOutput = this.modelsWithJSONOutput().includes(this.generativeModel() || '');
+        if (structredOutput) {
+          this.jsonOutputControl.enable();
+          this.isJsonOutputDisabled = false;
+        } else {
+          this.jsonOutputControl.patchValue(false);
+          this.jsonOutputControl.disable();
+          this.isJsonOutputDisabled = true;
```
ykulynych: I see that the isJsonOutputDisabled is just an attribute here. Since we use it in the template, I think it is better to convert it to the signal to prevent possible issues.

ykulynych (approved):

libs/common/src/lib/search-widget/search-configuration/results-display-form/results-display-form.component.ts
operramon: ops, added a new commit to fix it

PR #2884, opened 2026-06-05
Georgekulynych/sc 14814/manage structured field schemas

ykulynych:
image
image
image


libs/common/src/lib/knowledge-box-settings/kv-schemas/schema-form/schema-form.component.scss
```diff
@@ -0,0 +1,99 @@
...
+.form-actions {
+  display: flex;
+  justify-content: flex-start;
+  gap: rhythm(1);
+  margin-top: rhythm(4);
+  padding-top: rhythm(3);
+}
+
+.error-text {
+  color: var(--color-error-regular);
```
operramon: not important: it seems this color does not exist

operramon (approved):

libs/common/src/lib/knowledge-box-settings/kv-schemas/schema-form/schema-form.component.scss
ykulynych: Yes, fixed that

PR #2889, opened 2026-06-10
[sc-14889] Widget: Improved WCAG accessibility

ykulynych:


ebrehault (dismissed):

libs/search-widget/src/common/focusTrap.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    if (event.shiftKey) {
+      // Shift+Tab: if focus is on (or before) the first element, wrap to last
+      if (document.activeElement === first || !container.contains(document.activeElement)) {
+        event.preventDefault();
+        last.focus();
+      }
+    } else {
+      // Tab: if focus is on (or after) the last element, wrap to first
+      if (document.activeElement === last || !container.contains(document.activeElement)) {
```
operramon: A minor detail: while testing I noticed that the focus does not return to the first element.
It appears to happen when the widget is embedded as a custom element, not when using the `search-widget-demo`, so maybe it's related to the shadow dom.

operramon (dismissed):

libs/search-widget/src/common/focusTrap.ts
ykulynych: Great catch, i've fixed it

ebrehault (approved):

PR #2891, opened 2026-06-10
[sc-14888] Context box: Improved WCAG accessibility

ykulynych:


libs/common/src/lib/base/base.component.html
```diff
@@ -1,8 +1,17 @@
+<a
+  href="#dashboard-main"
```
operramon, line 2: It seems this link is not working in Angular

ykulynych: Fixed it, but for now its not very useful, because we see main block immediately. Hovewer, we could tune it later.

ebrehault (approved):

PR #2898, opened 2026-06-15
[sc-14945] Added 'cloud zero' and 'manual' subscriptions forms

ykulynych:


apps/manager-v2/src/app/manage-accounts/account-details/subscriptions/subscriptions.component.html
```diff
@@ -0,0 +1,164 @@
...
+      label="Billing formula"
+      formControlName="billing_formula_id"
+      [options]="service.formulasOptions | async"></pa-select>
+    <pa-input
+      formControlName="free_tokens_per_billing_cycle"
+      type="number">
+      Free tokens per billing cycle
+    </pa-input>
+    <pa-input
+      formControlName="free_storage_nuclia_tokens"
```
operramon, line 27: While testing, I noticed that when creating a manual subscription, the `free_storage_nuclia_tokens` property is not stored. Could it be a backend issue ?

ykulynych: Yes, the backend will add nuclia_tokens_price and free_storage_nuclia_tokens later. One more thing, you wont be able to test cloud_zero subscription on staging, it's available only on production.

operramon (approved):

PR #2927, opened 2026-06-22
[sc-14909] Added a prototype of the ai chat assistant

ykulynych:


libs/common/src/lib/chat-advice/chat-advice.service.ts
```diff
@@ -0,0 +1,263 @@
...
+        const navFields = contextualPages.map((p) => `t/${p.id}`);
+        const chatHistory = previousMessages.slice(-6).map((message) => ({
+          author: message.role === 'user' ? Ask.Author.USER : Ask.Author.NUCLIA,
+          text: message.content,
+        }));
+
+        return this.http
+          .post(
+            this.advisorKbAskUrl,
+            {
```
ebrehault, line 64: we want here an `audit_metadata` param that will contain:
- `origin: 'ai-assistant'`
- `account: '<the-account-id>'`

libs/common/src/lib/chat-advice/chat-advice.service.ts
```diff
@@ -0,0 +1,263 @@
...
+            { responseType: 'text' },
+          )
+          .pipe(
+            map((rawText) => this.extractExplanation(rawText)),
+            switchMap((explanation) =>
+              this.http
+                .post(
+                  this.advisorKbAskUrl,
+                  {
+                    query: `User context: ${hint}\n\nUser question: "${userMessage}"\n\nExplanation to enrich with links:\n${explanation}`,
```
ebrehault, line 82: same

ebrehault (approved):

PR #2938, opened 2026-06-30
[sc-14776] Added few improvements to metrics-pages, better communicat…

ykulynych:
…ion about REMi scores

 - Metrics pages: Replaced infinite scroll with page-based pagination (previous/next arrows, configurable page size)
 - REMi scores: Normalized display from raw 1–5 scale to 0–100% with status labels (Good / Needs review / Poor)
 - Metrics pages: Empty state now shows a contextual message; when filters are active and yield no results, a "Reset filters" button is shown
 - REMi analytics page: Month selector replaced with a dropdown (consistent with other metrics pages)
 - Metrics filters: "Add condition" changed from icon-only button to icon + label button


libs/common/src/lib/metrics/usage-analytics-page/usage-analytics-page.service.ts
```diff
@@ -305,7 +322,7 @@ export class UsageAnalyticsPageService extends AbstractMetricsPageService<UsageA
 
     const remiScores = [answerRelevance, contentRelevance, groundedness].filter((v): v is number => v !== null);
     const remiScore = remiScores.length > 0 ? Math.min(...remiScores) : null;
-
+    const issue = this._deriveIssue(remiItem.answer, status, answerRelevance, contentRelevance, groundedness);
```
operramon: Minor thing: it seems this variable is not used.

libs/common/src/lib/metrics/usage-analytics-page/usage-analytics-page.service.ts
```diff
@@ -305,7 +322,7 @@ export class UsageAnalyticsPageService extends AbstractMetricsPageService<UsageA
 
     const remiScores = [answerRelevance, contentRelevance, groundedness].filter((v): v is number => v !== null);
     const remiScore = remiScores.length > 0 ? Math.min(...remiScores) : null;
```
operramon, line 325: Just for curiosity,
do you know why the minimum score is used instead of the average ?

operramon (dismissed):

libs/common/src/lib/metrics/metrics-page.component.ts
```diff
@@ -86,6 +121,11 @@ export class MetricsPageComponent {
     effect(() => {
       this.service.setItems(this.items());
     });
+    // Reset to first page only on full data reloads (month/filter change), not on append
```
operramon, line 124: while testing I noticed that the pagination is not reset when changing the month, is it expected ?

ykulynych: No, that's a bug, ill fix it

libs/common/src/lib/metrics/usage-analytics-page/usage-analytics-page.service.ts
ykulynych: I discussed it with Carles Onielfa and he suggested to use min score there. As for me, it also sounds more logical to use average, but i'm not an expert here

libs/common/src/lib/metrics/usage-analytics-page/usage-analytics-page.service.ts
ykulynych: Oh, yes, ill remove it 😳

operramon (approved):

PR #2949, opened 2026-07-09
[sc-14993] Reworked account navigation

ykulynych:


libs/common/src/lib/navbar/navbar.component.html
```diff
@@ -426,153 +426,5 @@
...
-                data-cy="go-to-billing"
-                class="nav-link"
-                [class.sub-item]="inBilling | async">
-                <pa-icon name="payment"></pa-icon>
-                <span class="nav-link-title">{{ 'navbar.subscription' | translate }}</span>
-              </a>
-            </div>
-          }
-          @if ((isSubscribed | async) && (inBilling | async)) {
-            <div class="nav-group">
```
operramon, line 524: Just a small thing:
the menu items for `billing/checkout`, `/billing/my-subscription`, `/billing/usage`, `/billing/history` were deleted.
Is there any way to access these pages now ?

(These pages are only supposed to be visible for accounts subscribed to Stripe, so it's a bit difficult to test)

operramon (dismissed):

libs/common/src/lib/navbar/navbar.component.html
ykulynych: Yes, you can test it by entering url manually. Ive added redirects from the old routes to the new ones: https://github.com/nuclia/frontend/blob/8964525a7b5b6382260a81016fd5dc16723f69b3/libs/common/src/lib/account/account.module.ts

operramon: I see, entering the URL works well,
as for the users, will they be able to access these pages ?

ykulynych: No, only account managers/owners

operramon: I don't see any button to access the `/billing/my-subscription`, `/billing/history` pages,
will owners be able to access them ?

ykulynych: Hmm, it seems that we missed that.. You saved us 😄

operramon (approved):

PR #2964, opened 2026-07-20
Georgekulynych/sc 15095/agentic in rag agentic option in the search

ykulynych:


libs/common/src/assets/i18n/ca.json
```diff
@@ -1946,19 +1946,11 @@
 	"retrieval-agents.workflow.sidebar.test.description.ask": "Recuperant dades de la(es) font(s) seleccionada(es)...",
 	"retrieval-agents.workflow.sidebar.test.description.basic_ask": "Recuperant dades de la(es) font(s) seleccionada(es)...",
 	"retrieval-agents.workflow.sidebar.test.description.conditional": "Avaluació de la condició en funció del prompt…",
-	"retrieval-agents.workflow.sidebar.test.description.cypher": "",
 	"retrieval-agents.workflow.sidebar.test.description.data_viz": "Generant visualització de dades…",
-	"retrieval-agents.workflow.sidebar.test.description.external": "",
-	"retrieval-agents.workflow.sidebar.test.description.generate": "",
-	"retrieval-agents.workflow.sidebar.test.description.historical": "",
 	"retrieval-agents.workflow.sidebar.test.description.internet": "Recuperant dades d'Internet…",
-	"retrieval-agents.workflow.sidebar.test.description.mcp": "",
```
operramon, line 1955: These empty texts are strange, but I think they were left empty on purpose.

It is a way to avoid displaying a description for specific agents in this template:

https://github.com/nuclia/frontend/blob/2fcc2f0eac1a7d0da5647ff6924f62d545972789/libs/common/src/lib/retrieval-agent/agent-dashboard/workflow/sidebar/test-agent/elements/agent-block/agent-block.component.ts#L80

libs/common/src/lib/search-widget/search-configuration/search-configuration.component.html
```diff
@@ -120,59 +120,149 @@
...
     } @else if ((savedConfig | isTypedConfig) && (currentConfig | isTypedConfig)) {
+      <div class="search-mode-container">
+        <pa-radio-group
+          noBackground
+          [value]="searchMode()"
+          (valueChange)="updateSearchMode($event)">
+          <pa-radio value="agentic">
+            <span class="search-mode-label">
+              <span class="title-m">{{ 'search.configuration.search-mode.agentic' | translate }}</span>
+              <pa-button
```
operramon, line 131: While testing I noticed that these pa-buttons are not visible

libs/common/src/lib/search-widget/search-configuration/search-configuration.component.scss
```diff
@@ -57,6 +57,37 @@
...
+  }
+
+  .agentic-config-flow,
+  .rao-config-flow {
+    display: flex;
+    flex-direction: column;
+    gap: rhythm(1.5);
+  }
+
+  .rao-no-source {
```
operramon: it seems that classes like `rao-no-source` or `rao-config-flow` don't exist in the templates.

libs/common/src/lib/search-widget/connections.service.ts
operramon, line 1: Is this service used somewhere ?

operramon (approved):

libs/common/src/assets/i18n/ca.json
ykulynych: Oh.. i see, fixed that in a different way

libs/common/src/lib/search-widget/search-configuration/search-configuration.component.html
ykulynych: Good catch, imported needed modules

libs/common/src/lib/search-widget/search-configuration/search-configuration.component.scss
ykulynych: Removed them

libs/common/src/lib/search-widget/connections.service.ts
ykulynych: Well, i planned to use it later, but its definetely not needed here

operramon (approved):

PR #2976, opened 2026-07-22
[sc-15052] Display Key value fields in resources list

operramon:


libs/common/src/lib/resources/resource-list/resources-table/resources-table.component.html
```diff
@@ -298,6 +308,40 @@
                   }
                 </pa-table-cell>
               }
+              @if (visibleColumns.includes('key-value-fields')) {
+                <pa-table-cell>
+                  @let fields = row.keyValue || {} | keyvalue;
```
ykulynych, line 313: Just an idea - to format fields and date field if user will click 'Display full key-value' to show this
Image
instead of what we have now
Image

ykulynych (dismissed):

libs/common/src/lib/resources/resource-list/resources-table/resources-table.component.html
operramon: yes, looks good

ebrehault (approved):

PR #3003, opened 2026-07-31
[sc-15215] Create agentic config together with the agentic search config

ykulynych:


libs/common/src/lib/search-widget/search-configuration/agentic-configuration/agentic-configuration.component.html
```diff
@@ -1,85 +1,205 @@
...
+  <div class="config-form-field agentic-sources-field">
+    <span class="sources-label extra-label">{{ 'search.configuration.agentic.sources-label' | translate }}</span>
+    @if (hasAvailableSources()) {
       <div class="agentic-sources-list">
-        @for (source of sources(); track source.id) {
+        @for (sourceId of selectedSourceIds(); track $index; let rowIndex = $index) {
           <div class="agentic-source-row">
-            <span class="body-m">{{ source.id }}</span>
-            <nsi-badge kind="neutral">{{ connectionTypeLabels[source.type] ?? source.type }}</nsi-badge>
+            <pa-select
```
operramon, line 8: While testing I noticed that clicking the text of an option does not select the option.

libs/common/src/lib/search-widget/search-configuration/agentic-configuration/agentic-configuration.component.html
```diff
@@ -1,85 +1,205 @@
...
+        modalTitle="search.configuration.agentic.new-source.description-label"
+        (valueChange)="updateNewSourceDescription($event)"
+        (resizing)="heightChanged.emit()">
+        {{ 'search.configuration.agentic.new-source.description-label' | translate }}
+      </nsi-expandable-textarea>
+    }
+  </div>
+
+  <div class="config-form-field">
+    <pa-select
```
operramon, line 64: Maybe there's a backward compatibility issue: if a user has a previously created search configuration with `generativeAnswer.generateAnswer: false`, the new UI will display this configuration as `simple-rag` mode. Should it be `search` mode ?

libs/common/src/lib/search-widget/search-widget-storage.service.ts
```diff
@@ -226,6 +248,47 @@ export class SearchWidgetStorageService {
...
+        ),
+      ),
+    );
+  }
+
+  private _deleteAgenticConfig(id: string) {
+    // Best-effort cleanup: a 404 (config never existed for this widget config name) is expected and ignored.
+    return this.sdk.currentKb.pipe(
+      take(1),
+      switchMap((kb) => kb.deleteAgenticConfig(id).pipe(catchError(() => of(undefined)))),
```
operramon, line 276: It seems that when calling this endpoint, it returns `405 (Method Not Allowed)`

ykulynych: Yes, thats expected, waiting when backend team will allow to delete agentic configs

libs/common/src/lib/search-widget/search-configuration/agentic-configuration/agentic-configuration.component.html
ykulynych: Good catch, fixed that

libs/common/src/lib/search-widget/search-configuration/agentic-configuration/agentic-configuration.component.html
ykulynych: My bad, fixed this issue

operramon (approved):

PR #3010, opened 2026-08-04
[sc-15249] Improve performance in resource page

operramon:


libs/common/src/lib/resources/edit-resource/edit-resource.service.ts
```diff
@@ -120,7 +121,28 @@ export class EditResourceService {
...
+            ResourceProperties.SECURITY,
+          ])
+          .pipe(
+            switchMap((resource) => {
+              // If all fields are conversations, we don't need to retrieve extracted data.
+              // This is a temporary workaround to prevent memory resources from taking too long to load.
+              const onlyConversations = Object.keys(resource.data).every((fieldType) =>
+                ['conversations', 'generics'].includes(fieldType),
+              );
+              return onlyConversations ? of(resource) : kb.getFullResource(resourceId);
```
lferran: We can't assume that all fields will be conversations when a KB has memory data. It can also have content fields.

The idea would that "getFullResource" does not load the field values or extracted data. These are retrieved only if the user clicks on the field and visits the field page.

ykulynych (dismissed):

libs/common/src/lib/resources/edit-resource/edit-resource.service.ts
operramon: Fixed: now it works for any type of fields

ykulynych (dismissed):

ykulynych (approved):

PR #3021, opened 2026-08-07
[sc-15227] Redesigned consumption page

ykulynych:
image


libs/common/src/lib/account/account-home/account-consumption.component.ts
```diff
@@ -22,14 +45,44 @@ export class AccountConsumptionComponent implements OnInit, OnDestroy {
...
   usage?: { [key: string]: UsagePoint[] };
   tokensCount?: { [key: string]: number };
+  accountTokens = 0;
 
   kbs = this.sdk.kbList;
   nuaKeys = this.sdk.currentAccount.pipe(
     switchMap((account) => this.sdk.nuclia.db.getNUAClients(account.id)),
     shareReplay(1),
   );
+  isNuaActivityEnabled = this.features.unstable.viewNuaActivity;
```
operramon, line 55: just a minor thing, I just remembered that the activity page for NUA is obsolete. 
I guess we should have removed it long ago.

operramon (approved):

PR #3027, opened 2026-08-13
[sc-15241] Show trial info, plan name, removed upgrade button for pro…

ykulynych:
…/growth plans


apps/dashboard/src/app/knowledge-box/knowledge-box-home/knowledge-box-home.component.html
operramon: One small detail: the `app-account-status` component includes an additional feature that checks if the trial period has ended and displays a modal: https://github.com/nuclia/frontend/blob/main/libs/common/src/lib/account/account-status/account-status.component.ts#L75

After this commit, this additional feature will not be active because the `app-account-status` component is no longer used anywhere.
Do you know we want to remove this feature ?

operramon (approved):

PR #3029, opened 2026-08-17
[sc-15288] Implemented progressive loading of the usage data

ykulynych:


libs/sdk-core/src/lib/db/db.spec.ts
```diff
@@ -42,6 +42,7 @@ describe('Db', () => {
         id: 'qwerty',
         slug: 'geb',
         zone: '1',
+        zone_id: 'qwerty',
```
operramon, line 45: does the `zone_id` parameter exist in a kb ?

operramon (approved):

libs/sdk-core/src/lib/db/db.spec.ts
ykulynych: no, but it needed by /index/kbs endpoint

operramon: ah, makes sense

operramon (approved):

PR #3047, opened 2026-08-26
feat(retrieval-agents): smarter workflow landing + workflows list polish

dowsprogress:
## Summary

Three related UX improvements to the Retrieval Agents workflows experience: a canvas back button, smarter landing behavior when entering an agent, and a decluttered workflows list header/actions.

---

### 1. Back button in the canvas toolbar

**What changed:** Added a dedicated back button (chevrons-left icon, with a "Back to workflows list" tooltip) to the workflow canvas toolbar, positioned between the nav toggle and the workflow picker. Clicking it navigates back to the workflows list.

**UX justification:** The canvas and the workflows list are router siblings, but without an explicit way back the canvas felt like a disconnected, dead-end page — the only way out was browser back or the nav sidebar. A visible, always-available back button makes the relationship between "list of workflows" and "a workflow's canvas" obvious and removes reliance on browser navigation.

### 2. Land on the default or last-visited workflow canvas

**What changed:** Entering an agent used to always redirect to the workflows *list*. It now resolves, via a route guard, to the workflow the user last worked on (persisted per-agent in `localStorage`), falling back to the seeded `default` workflow, and only falling back to the list if neither exists.

**UX justification:** Almost every agent has at least one workflow (all agents ship with a `default` one), and most users work in a single workflow most of the time. Forcing a list-page stop on every entry added a redundant click for the common case and broke the "resume where I left off" expectation. This doesn't regress the back button above (#1), which still explicitly navigates to `/workflows` on demand.

### 3. "Open workflow" menu option + redirect to canvas after creation

**What changed:** Added an "Open workflow" item to each row's overflow (⋮) menu, and after creating a workflow from the "Add workflow" modal, the user is now taken straight to its canvas instead of staying on the list.

**UX justification:** Rows were already clickable to open the canvas, but that wasn't obvious — no visual affordance signaled it. Rather than adding a new chevron/link treatment, we reused the existing overflow menu users already open for Edit/Delete, making the action discoverable without adding UI clutter. Separately, since the entire point of creating a workflow is to start building it, staying on the list after creation was an unnecessary extra click.

### 4. Primary "Add workflow" button + MCP endpoint moved into a modal

**What changed:**
- "Add workflow" is now styled `kind="primary"` (blue), matching the primary-action pattern used elsewhere in the dashboard (e.g. the Knowledge Box header).
- The MCP endpoint block — previously a permanently visible code snippet + copy button taking up page real estate — was moved into a small modal, accessed via a new overflow icon button next to "Add workflow". The modal mirrors the existing "Developer integrations" copyable-endpoint modal pattern already used on the dashboard home.

**UX justification:** Having the Add button visually equal-weighted with the (unrelated) endpoint copy action diluted the page's single clear call-to-action. The MCP endpoint is a power-user/integration detail needed occasionally, not on every visit — hiding it behind one extra click declutters the default view while keeping it fully accessible and consistent with an established pattern elsewhere in the product.

---

## Screenshots / how to verify
- Canvas toolbar → back button (chevrons-left) next to the nav toggle → hover shows "Back to workflows list" tooltip → click returns to the workflows list.
- Enter an agent that has a `default` workflow (or one you've visited before) → you should land directly on its canvas.
- Workflows list: "Add workflow" is blue; new overflow icon (⋮) next to it opens "MCP endpoint" → modal with copyable field.
- Row overflow menu → "Open workflow" navigates to that workflow's canvas.
- Create a new workflow → lands on its canvas immediately.

## Validation
- `nx lint common` — clean (only pre-existing unrelated warnings)
- `nx build dashboard` — succeeds (only pre-existing unrelated warnings)
- `nx test common` — 198/199 pass (1 pre-existing unrelated failure, date-based)

## Notes
- No backend/API changes.
- A previously-investigated "can't recreate a workflow with the same name right after deleting it" issue was intentionally **not** addressed in this PR — the fix attempted (`synchronous: true` on write calls) didn't fully resolve it on further testing and was reverted; root cause looks backend-side and is out of scope here.

### DIFF
--- a/libs/common/src/lib/retrieval-agent/agent-dashboard/agent-dashboard.component.ts
+++ b/libs/common/src/lib/retrieval-agent/agent-dashboard/agent-dashboard.component.ts
@@ -11,7 +11,13 @@ import {
   ViewEncapsulation,
 } from '@angular/core';
 import { ActivatedRoute, Router } from '@angular/router';
-import { ModalConfig, PaButtonModule, PaDropdownModule, PaPopupModule } from '@guillotinaweb/pastanaga-angular';
+import {
+  ModalConfig,
+  PaButtonModule,
+  PaDropdownModule,
+  PaPopupModule,
+  PaTooltipModule,
+} from '@guillotinaweb/pastanaga-angular';
 import { TranslateModule } from '@ngx-translate/core';
 import { auditTime, combineLatest, filter, fromEvent, map, Subject, switchMap, take, takeUntil } from 'rxjs';
 import { DashboardLayoutService } from '../../base';
@@ -53,6 +59,7 @@ const OVERFLOW_SIDEBARS: SidebarType[] = ['rules', 'import', 'export'];
     PaButtonModule,
     PaDropdownModule,
     PaPopupModule,
+    PaTooltipModule,
     TranslateModule,
     WorkflowRootComponent,
   ],
@@ -176,6 +183,10 @@ export class AgentDashboardComponent implements AfterViewInit, OnDestroy {
 
   goToSessions() {
     this.router.navigate([`${this.aragUrl()}/sessions`]);
+  }
+
+  backToWorkflows() {
+    this.router.navigate([`${this.aragUrl()}/workflows`]);
   }
 
   addWorkflow() {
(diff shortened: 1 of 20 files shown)


libs/common/src/lib/retrieval-agent/agent-dashboard/agent-dashboard.component.ts
operramon, line 1: It seems there's a git conflict in this file, I can look into it later

dowsprogress: thanks

operramon: The conflicts are resolved,
but maybe there's a bug when switching from one retrieval agent to another. I'll investigate...

operramon: I noticed a few bugs when switching between Retrieval agents using the top bar selector. I've added a commit to fix them. It should work fine now.

operramon (approved):

PR #3050, opened 2026-08-26
Georgekulynych/sc 15137/pendo integration

ykulynych:


apps/dashboard/src/index.html
ebrehault, line 1: we also need it in the auth app

ebrehault (changes requested):

apps/dashboard/src/index.html
ykulynych: what should we use for visitor_id and account_id there?

ebrehault (approved):

PR #3065, opened 2026-09-01
[sc-15342] Support extra params on Perplexity and Google sources

operramon:


libs/sync/src/lib/add-source-page/google-source/google-source.component.html
```diff
@@ -0,0 +1,20 @@
+<div
+  class="extra-form"
+  [formGroup]="form">
+  <nsi-expandable-textarea
```
ebrehault, line 4: we need a placeholder or an help text to explain the expected format (like `en.wikipedia.org`)

libs/sync/src/lib/add-source-page/perplexity-source/perplexity-source.component.html
```diff
@@ -0,0 +1,11 @@
+<div
+  class="extra-form"
+  [formGroup]="form">
+  <nsi-expandable-textarea
```
ebrehault, line 4: we need a placeholder or an help text to explain the expected format (like `en.wikipedia.org`)

ebrehault (dismissed):

libs/sync/src/lib/add-source-page/google-source/google-source.component.html
operramon: Fixed!

ebrehault (approved):

PR #3082, opened 2026-09-09
feat(dashboard): streamline app header controls

dowsprogress:
## Summary

Streamlines the dashboard app header by separating **application context** from
**personal controls**:

- moves the Knowledge Box picker beside the product logo
- restyles the picker as a quiet, content-sized context switcher
- consolidates notifications into the avatar menu
- replaces the decorative settings badge with a meaningful unread indicator
- uses the current account name as the account-menu heading
- keeps the KB storage information trigger visible while its data loads

### Reduce persistent clutter and cognitive load

The previous app bar presented the account name, KB picker, plan usage, notification
bell, settings-marked avatar, and page controls at the same time. Even though each
element was individually understandable, their combined visual weight created a
dense strip of persistent chrome that users had to repeatedly scan and interpret.

This change reduces the number of always-visible controls and gives each remaining
element a clearer role. Context switching moves beside the product mark,
notifications become visible on demand with an ambient unread signal, and account
identity is consolidated into the avatar menu. The result uses progressive
disclosure to preserve every capability while lowering visual competition and
leaving more attention for the page's primary task.

Reducing this top-level clutter is especially important because the app bar appears
across the product: small sources of ambiguity and cognitive overhead are repeated on
every page and every navigation.

### Put context on the left and identity on the right

The previous header mixed three different types of information in one right-aligned
cluster: the active Knowledge Box, plan usage, and the user's account menu. This made
the hierarchy unclear because a high-frequency context switcher looked like another
personal or administrative control.

The new layout follows the established application-header pattern used by products
such as Vercel and GitHub:

```text
Product / Current context ▾                         Plan usage   Avatar
```

The active Knowledge Box now sits beside the product mark because it answers
**"where am I working?"**. The avatar stays on the far right because it answers
**"who am I?"**. This is not intended as a breadcrumb: the slash is a lightweight
visual separator between product and context, not a representation of navigational
hierarchy.

When there is only one available KB or retrieval agent, the name remains visible as
a static context label while the dropdown affordance is removed. This avoids a
redundant control without losing the user's orientation in the product.

### Make the picker quieter without hiding its affordance

The existing `nsi-dropdown-button` is designed primarily for forms and carries a
fixed minimum width, outlined treatment, semibold label, and generous padding. Those
properties made the current KB look like a primary action.

The topbar applies a scoped, intentional exception:

- sizes the picker to its content
- removes the outline
- uses regular label weight
- retains the chevron and hover background as the interaction cues
- removes the persistent lock icon from the trigger to reduce noise

Privacy remains visible inside the dropdown list, where it is useful for comparing
available contexts.

This matches the industry-standard workspace/project switcher pattern: persistent
context is legible, but visually subordinate to page-level actions.

### Consolidate notifications into the account menu

The standalone notification bell consumed permanent header space even when there
were no unread items. Notifications now live directly below the current account name
inside the avatar menu:

```text
ACCOUNT NAME
Notifications                                      3
Consumption
Account settings
```

Selecting the row opens the existing notification side panel, so the destination and
interaction model are unchanged. The exact unread count is right-aligned in the menu
row, while a small dot on the avatar preserves ambient awareness when the menu is
closed.

The dot replaces the previous green gear badge, which looked interactive but had no
independent action. This gives that visual location meaningful status rather than
decoration.

The avatar trigger also exposes the unread count in its accessible label, so the dot
is not the only notification signal.

### Make account context consistently visible

The old account label only appeared in the topbar when a user belonged to multiple
accounts, while the generic `Account` menu heading only appeared for account
managers. The menu now uses the actual account name as its heading for every user.

This improves orientation without widening permissions: account-management items
retain their existing visibility rules.

The heading uses the primary brand colour to distinguish the current account context
from generic menu section labels.

### Preserve the KB storage affordance during loading

The storage-summary info icon beside the page-level KB title was conditionally
removed whenever counters were temporarily unavailable. This could make the control
appear to vanish during loading.

The trigger now remains structurally present and is disabled until the counters are
available. Once loaded, it opens the same storage-summary dropdown as before.

## Implementation notes

- The topbar is shared by the dashboard and RAO applications.
- The quiet picker styling is deliberately scoped to the topbar rather than changing
  the shared `nsi-dropdown-button` defaults and potentially reflowing form controls
  elsewhere.
- The removed `NotificationButtonComponent` had no remaining consumers.
- Notification behaviour in simple mode remains unchanged: the menu item is hidden
  wherever the previous bell was hidden.
- Standalone mode continues to use `app-standalone-menu` and is unaffected.

## Validation

- Dashboard production build
- RAO production build
- Dashboard KB-header component tests: 6/6
- Pastanaga dropdown and option tests: 8/8 during implementation
- Affected lint: 0 errors
- Prettier and `git diff --check`

## Review checklist

- [ ] Verify the logo / KB picker spacing and hover treatment in light and dark modes
- [ ] Verify a single context renders as a static name with no chevron or popup
- [ ] Verify long KB names do not collide with plan usage or avatar controls
- [ ] Verify unread and zero-unread avatar states
- [ ] Verify notification count alignment and `9+` overflow in the account menu
- [ ] Verify the Notifications row opens and the panel closes normally
- [ ] Verify account-manager and non-manager menu variants
- [ ] Verify the KB storage info icon while counters are loading and after they load


libs/common/src/lib/topbar/kb-switch/kb-switch.component.html
```diff
@@ -1,14 +1,20 @@
...
+    <nsi-dropdown-button
+      size="small"
+      aspect="basic"
+      freeWidth
+      openOnFocus
+      [popupRef]="kbSwitch">
+      {{ current.title }}
+    </nsi-dropdown-button>
+  } @else {
+    <span
```
operramon, line 12: Hi, while testing I noticed two minor things:
- When an account has only one KB, the ellipsis doesn't seems to work
- In the "Context Box" product, should the KB title be visible ?

dowsprogress: **-When an account has only one KB, the ellipsis doesn't seems to work** - this is intended it should just display the current KB as there is nothing to change to
**- In the "Context Box" product, should the KB title be visible ?** - no, it does not currently on live app as far as I'm aware? so behaviour should remain the same

operramon: ok 👍, I have added a small commit to hide the KB title in Context Box.
Also I fixed a few git conflicts

operramon (approved):

PR #3099, opened 2026-09-14
Upgrade to angular 22

mpellerin42:
Update nx to version 23.1 (including angular 22)


apps/admin/src/app/account/account-nua/nua-activity/nua-activity.component.html
```diff
@@ -7,7 +7,7 @@
         routerLink="../..">
         {{ 'generic.back' | translate }}
       </pa-button>
-      <h1>{{ 'account.nua_key_activity_title' | translate: { key: (client | async)?.title } }}</h1>
+      <h1>{{ 'account.nua_key_activity_title' | translate: { key: $safeNavigationMigration((client | async)?.title) } }}</h1>
```
operramon, line 10: just out of curiosity,
whenever we find a `$safeNavigationMigration`,
can we remove it if `null` or `undefined` has the same effect on the template ?

operramon (approved):

apps/admin/src/app/account/account-nua/nua-activity/nua-activity.component.html
mpellerin42: I would say yes, this was part of the automatic Angular migration steps. I kept them all so we can then take the time to review in dept what they do and if we can remove them, that's good. 
Same for all the EAGER strategy which are now everywhere, I think at some point we will review as EAGER is not the way we want to go, but as it was the default for so long it might not be straightforward to remove them all 😅

PR #3105, opened 2026-09-15
Hide the version number in the dashboard

operramon:


libs/common/src/lib/navbar/navbar.component.ts
```diff
@@ -77,9 +77,9 @@ export class NavbarComponent implements OnInit, OnDestroy {
   isAragAdmin = this.features.isAragAdmin;
   isAccountManager = this.features.isAccountManager;
   noStripe = this.backendConfig.noStripe();
-  version = this.backendConfig.getVersion();
+  versionHash = this.backendConfig.getVersion().replace(/^\d+\.\d+\.\d+-/, '');
```
mpellerin42: I feel like this kind of computation should be done directly in backend-config service so it can be reused elsewhere if needed (I'm not sure all our apps are using the navbar).
Like either refactor getVersion() method to return only the hash, or add a getVersionHash() which would return only the hash.

operramon: ok, it makes sense, it's fixed now.

mpellerin42 (approved):

PR #3111, opened 2026-09-16
Modernize boot splash screen across all apps

dowsprogress:
## What
Replaces the dated 2015-era spinner + hardcoded emoji copy on the pre-bootstrap boot splash (shown in `dashboard`, `auth`, `admin`, `rao` before Angular mounts) with a pulsing app logo, and fixes a real white-screen flash bug in the hide timing.

## Changes
- **Removed**: old spinner CSS/markup + hardcoded emoji copy
- **Added**: pulsing logo (`logo-horizontal.svg`, the same default top-bar logo), using Tailwind's industry-standard `animate-pulse` spec (2s, `cubic-bezier(0.4,0,0.6,1)`, opacity 1 to 0.5), with `prefers-reduced-motion` support
- **Fixed** (real bug, not cosmetic): splash was being hidden directly from each app's `app.component.ts` `ngAfterViewInit()`, firing before routing/paint completed and causing a white flash. Hide logic is now centralized in `STFSplashScreenService`, gated on router `NavigationEnd` + two chained `requestAnimationFrame` calls (waits for one real paint cycle without relying on `ApplicationRef.isStable`, which doesn't reliably settle in this app)

## Verification
- `nx lint` clean (0 errors) on `dashboard`, `auth`, `admin`, `rao`, `core`
- Manually verified via local dev server: logo renders, pulses, and splash hides cleanly with no white flash
- Pre-existing, unrelated production bundle-budget failure on `nx build dashboard --configuration=local-stage` confirmed unaffected by this change (dev-mode bundle size issue, not caused by this PR)

## For the implementing dev
Nothing else to wire up — this is self-contained CSS/HTML + one service + a one-line DI touch per app. No config, flags, or backend changes needed.


mpellerin42 (approved):
Nice !
I just moved the splash-screen css to sistema so all the apps use the same one instead of duplicating it everywhere

PR #3100, opened 2026-09-14
[UX] NUA Guard policy management UI

dowsprogress:
Relates to #3096

## Summary
Implements the prototype UI for account-level NUA Guard query moderation policies (nuclia/frontend#3096).

## Changes
- Added `AccountNuaGuardComponent` list page and `PolicyDialogComponent` modal in `apps/admin/src/app/account/account-nua-guard/`.
- Added route `configuration/nua-guard` and tab to `AccountConfigurationComponent`.
- Added NUA Guard item to avatar menu under Configuration.
- Added i18n keys across `en`, `es`, `fr`, and `ca`.
- Created mock CRUD service with 5-enabled policy limit.

## Validation
- `nx build admin` and `nx build dashboard` pass cleanly.
- Verified in light and dark modes.

Issue #3096, opened 2026-09-14
[UX] NUA Guard


ebrehault:
NUA Guard evaluates user queries against natural-language policies before they are sent to a generative model.
It is based on policies that are defined at account level but attached to a given region (The policy will only apply to KBs inside this region).

In the account management pages, we need a way to manage our NUA Guard policies.
It is structured by region (if the current account has KBs only in a unique region, no need to show the region).

The needed fields are:

| Field | Description |
| --- | --- |
| `name` | A name used to identify the policy. |
| `description` | An optional customer-facing explanation returned when the policy blocks a request. |
| `instruction` | The moderation context, expected behavior, and desired strictness. |
| `query` | One yes/no question about the user content. |
| `enabled` | Whether the policy is evaluated automatically on chat requests. |
| `blocking` | Whether a violation or evaluation failure prevents generation. |

Example payload:
```json
{
  "name": "Financial advice",
  "description": "Requests for personalized financial advice are not allowed.",
  "instruction": "Review user requests for prohibited financial advice. Apply the policy strictly.",
  "query": "Does this request ask for personalized investment advice?",
  "target": "QUERY",
  "enabled": false,
  "blocking": true
}
```

Note:
- Up to five enabled policies per account

PR #3123, opened 2026-09-21
[gh-3096] nua guards

mpellerin42:
Resolves to https://github.com/nuclia/frontend/issues/3096.

Based on UX work done on https://github.com/nuclia/frontend/pull/3100


apps/admin/src/app/account/account-configuration/account-configuration.component.html
operramon, line 1: While testing, I created a NUA guard and then I tried to make a forbidden question using the `/predict/chat` endpoint with a NUA key. But it appears that the questions was not blocked as I expected.
is the feature flag added because the backend is not ready ?

operramon (approved):

apps/admin/src/app/account/account-configuration/account-configuration.component.html
mpellerin42: Did you enable the NUA guard before testing? If so, it's something to double check with Carmen, as the feature flag should not block Stage.

operramon: yes, the "enabled" toggle is enabled

